PR #2, opened 2023-05-16
docs: Add docs (build)

sandrina-p:
This library was part of [`@remotecom`](https://github.com/remoteoss) internal source code.

As we are open-sourcing it, this is the docs (build version). 

The library code itself is copied at #1 .

### Why not open-sourcing the docs' source?

Although the library code is open-sourced with unit tests, the docs cannot be moved yet. Our docs are still coupled to Remote's internal Design System and integration tests. The effort to decouple it at the moment is too high.

So our strategy is to at each new PR, to update the docs internally and manually copy its build to this repo. 

### How to review this
- Check the website where the docs are hosted (Vercel) is working well

PR #3, opened 2023-05-16
build: Scripts to manage release process

sandrina-p:
Added the scripts that we must run to publish a new release.

- `npm run release:dev:patch` -> To publish dev releases in PRs
- `npm run release:main:patch` -> To publish stable releases in main.

Updated the `CONTRIBUTING.md` explaining this development flow.

- Loom: [Development overview](https://www.loom.com/share/34554039380648b1b2922333d4e4b51a) - basically I do the steps written in the CONTRIBUTING file. 
- Loom: [scripts walkthrough](https://www.loom.com/share/4955fe2845d542dda327d8d9565e64a0) - a detailed overview of the scripts created.

Important:
- **DO NOT** test it in this repo, otherwise you'll publish unwanted versions. Use https://github.com/remoteoss/test-pkg-npm instead.
- Before you test, please copy the `scripts/release*` from here to there, to ensure you have the latest updates.


CONTRIBUTING.md
```diff
@@ -4,30 +4,86 @@
 
 If you have questions about the library, found a bug or want to suggest a feature, please create an issue.
 
-## Development
+## Documentation
 
-- Install dependencies:
+You can visit the [docs website](https://json-schema-form.vercel.app/), however its source is not in the repo yet, only the build (`/docs_build`). Our docs are still coupled to Remote's internal Design System and integration tests. The effort to decouple it at the moment is too high.
```
brennj, line 9: Glad we're being explicit about this

scripts/release_dev.js
```diff
@@ -1,62 +1,55 @@
...
+  const result = await execSync('git rev-parse --short HEAD^');
+  const gitSHA = result.stdout.toString().trim();
+  const currentVersion = packageJson.version;
 
-function getNewVersion() {
-  if (packageJson.version.includes('-dev.')) {
+  if (currentVersion.includes('-dev.')) {
     console.log('Bumping exisiting dev...');
-    return packageJson.version.replace(/-dev.\d{15}$/, `-dev.${currentDate}`);
+    return currentVersion.replace(/-dev\.(.*)$/, `-dev.${gitSHA}`);
```
brennj, line 24: <3

scripts/release_main.js
```diff
@@ -81,8 +97,9 @@ async function init() {
   );
 
   if (answerChangelog === 'no') {
+    // TODO revertCommit() to do this automatically.
     console.log('You may want to revert the changed files!');
-    process.exit();
+    process.exit(1);
   }
 
   await commit({ newVersion });
```
brennj, line 105: dumb q. how do you protect `main` when you're doing this?

sandrina-p: We don't. It's shitty. Ideally, this should happen after a feature PR is merged:
- GH Action triggers, which opens a new MR with the new version + CHANGELOG.
- You (or any maintainer) review the MR and approve it. Or tweak it if needed, or even close it, in case it doesn't make sense to release a new version.
  - On merge, another action is triggered to do the actual release.

But as I don't know Github Actions I couldn't make it happen :(

scripts/release_dev.js
```diff
@@ -1,62 +1,55 @@
...
-    pad2(date.getMinutes()) +
-    pad2(date.getSeconds())
-  );
-}
+const packageJsonPath = path.resolve(__dirname, '../package.json');
+const packageJson = require(packageJsonPath);
 
-const currentDate = getDateYyyyMMDDHHMMSS();
+async function getNewVersion() {
+  const result = await execSync('git rev-parse --short HEAD^');
```
brennj, line 19: do you still need the await if its `Sync`?

sandrina-p (commented):
@johnstonbl01 and @johnstonbl01 this is (finally) ready for review!

P.S. This is pointing to `copy-sourcecode` branch so it benefits from the eslint friends. I'll rebase it in a few minutes after that PR is merged.

scripts/release.helpers.js
```diff
@@ -0,0 +1,116 @@
...
+const { exec } = require('child-process-promise');
+
+async function checkGitStatus() {
+  console.log('Checking your PR branch...');
+  const resultBranch = await runExec('git branch --show-current', {
+    silent: true,
+  });
+  const branchName = resultBranch.stdout.toString().trim();
+  if (branchName === 'main') {
+    console.error(`🟠 You are at "main". Are you sure you wanna release a dev version here?`);
```
brennj, line 12: lots of good defensive programming here that we need

scripts/release_dev.js
sandrina-p: No, this was old code, it's gone

scripts/release.helpers.js
sandrina-p: This PR is the biggest proof I'm human. Each `if` was a mistake I've done. 😅

scripts/release_dev.js
```diff
@@ -0,0 +1,95 @@
+const path = require('path');
```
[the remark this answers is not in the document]
sandrina-p: Good point. Yes, because the branch exists in the origin. So it's still public/visible even though it doesn't has an explicit PR.

scripts/release_main.js
```diff
@@ -0,0 +1,130 @@
...
+  await checkGitStatus();
+}
+
+function getNewVersion() {
+  const currentVersion = packageJson.version;
+
+  // TODO later - get the right version based on commit history.
+  const versionType = process.argv.slice(2)[0];
+  const versionBase = semver.coerce(currentVersion);
+  return semver.inc(versionBase, versionType) + '-beta.0';
```
brennj, line 49: `beta.0` is maybe a little weird to me but I'm trying to get my thoughts correct here:

- We want to keep it unstable until we're ready for a correct `1.0.0`.
- So then wouldn't we be bumping the `beta.X` instead? and its `1.0.0.beta.123` until we're happy? 
- Thats not necessarily semver though... but for all patches, it should maybe do that instead?

sandrina-p: Honestly I'm not sure either, so here's my answer:

`beta.0` -> it's mandatory to have a number after any prefix. it cannot be just `.beta`

- My idea would be to release `0.1.0-beta.0` and then `0.x.x-beta.0` as we iterate. I'd prefer that over `.beta-x` so it's clear when new releases are features or fixes. Those this make sense?
- For that reason, I don't see us bumping `beta.x`. AFAIK `beta.x` is used to fix bugs directly related to that given version, not to fix bugs or release new feature in general.

Let me try to give an example. We start with `0.1.0-beta.0`.
- `0.2.0-beta.0` -> we now support this "foo1" thing.
- `0.2.0-beta.0` -> we now support this "bar" thing.
- `0.2.1-beta.0` -> we now fixed a bug at "biz" thing.

See, each commit it's about different features within the library as a whole. If we used `0.1.0-beta.x`, then it would be like:
- `0.2.0-beta.0` -> we now support this "foo1" thing.
- `0.2.0-beta.1` -> we now support this "foo2" thing.
- `0.2.0-beta.2` -> we now fixed a bug at "bar" thing.

It's not clear if the new versions are fixes, features, etc. 

In my understanding `beta` is used in the following scenario:
- `0.2.0-beta.0` -> we now support this "foo" thing.
- `0.2.0-beta.1` -> we fixed a bug related to the "foo" thing support this "foo2" thing.
- `0.2.0-beta.2` -> we fixed another bug related to the "foo" thing.
- `0.2.0-beta.3` -> we need to ship "bar" this. Does it make sense to be here? it should be "0.3.0-beta.0" instead.

Does this make sense? Honestly I don't know, I didn't find any document super clear about this, and looking at the NPM ecosystem around didn't bring me clarity either :/

brennj: > beta.0 -> it's mandatory to have a number after any prefix. it cannot be just .beta

TIL 💡 

My idea would be to release 0.1.0-beta.0 and then 0.x.x-beta.0 as we iterate. I'd prefer that over .beta-x so it's clear when new releases are features or fixes. Those this make sense?
For that reason, I don't see us bumping beta.x. AFAIK beta.x is used to fix bugs directly related to that given version, not to fix bugs or release new feature in general.

Thats perfectly fine then, lets roll with it!

sandrina-p: Cool! Let me just rewrite `main` to add the co-authors, then rebase #2, and finally come back to this one. Give me ~10min.

CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+
+You can create as many dev releases as you need during the PRs, running the same command.
+
+### Merging a PR
+
+A PR needs all CI checks to pass.
+
+By default, prefer to "Squash and Merge" giving it a message that follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
+
+The final release is done after merge.
```
brennj, line 70: Confirm for me:

It does not matter if a dev version gets into `main`, as long as it's followed up with the actual stable release. I don't need to necessarily revert any commits with the `package-lock.json`/`package.json` changes back to what the current stable version is.

brennj:
i'm happy with the steps as our first iteration, there is a lot of manual work - but this is ok for now. I will recheck later when its rebased. @johnstonbl01 defo will appreciate your eyes here too

CONTRIBUTING.md
brennj: perfect 4 me

scripts/release_dev.js
```diff
@@ -0,0 +1,109 @@
...
+    console.error(`🟠 You are at "main". Are you sure you wanna release a dev version here?`);
+    process.exit(1);
+  }
+
+  await checkGitStatus();
+}
+
+async function getNewVersion() {
+  const result = await runExec('git rev-parse --short HEAD^');
+  // BUG TODO later The gitSHA does not match the real next commit sha
```
johnstonbl01, line 33: I think this is because this command is returning the _next to last_ hash. 

I think `git describe --always` or `git log -1 --abbrev-commit --pretty=oneline` should get you the correct hash?

johnstonbl01:
I think there's a lot of opportunities here to automate some of this away, but for now having the manual processes is fine. 👍 Looks good to me, Sandrina!

scripts/release_dev.js
sandrina-p: Hum no. `git describe --always` returns the hash of the last commit. Here I want to get the hash of the next (upcoming not yet done) commit. :/

sandrina-p: Now that I have the approval of both John and Blake, shall we try to create `@0.1.0-dev.xxxx` in this PR or in another PR? 😅 I'd say let's go with this one.

johnstonbl01: > Hum no. `git describe --always` returns the hash of the last commit. Here I want to get the hash of the next (upcoming not yet done) commit. :/

Ah ok. It's not a blocker, but think we should land on something we want to keep for the long-term soon-ish.

sandrina-p: The alternative, which was my first commit, was to use date instead. eg `0.1.0-dev.20230517172340` (yyyymmddhhmmss). I've seen this in a couple of other libraries in the wild (eg [this one](https://www.npmjs.com/package/@pandacss/node?activeTab=versions)). WDYT?

johnstonbl01: 🤔 I kind of like the date + timestamp tbh.

sandrina-p: I agree. Sorry @brennj I'm turning back to the date again. https://github.com/remoteoss/json-schema-form/pull/3/commits/efe38a131727ea7cd7e2141ae52187d98d5d85a5 Yesterday you asked for the advantages, here's the answer:

- git sha do not match with the upcoming commit. On top of that, they are not historically sortable.
- By using the yyyymmddhhmmss format, we know their order of releases.

brennj: i'm ok with that ;)

sandrina-p (commented):
Check my latest commit. https://github.com/remoteoss/json-schema-form/pull/3/commits/28792ddde1d9b12a50c36829f96e9996c941f079 I noticed a "missing feature" where the actual package would be published empty. Why? Because it was missing running the `build` script to generate the `dist/` during the release script.

scripts/release_dev.js
```diff
@@ -0,0 +1,111 @@
...
+    process.exit(1);
+  }
+
+  // Check Auth/OTP before the bumpVersion() to reduce the probability
+  // of errors, leading files to be changed/pushed before the actual release.
+  await checkNpmAuth();
+  const otp = await askForText('🔐 What is the NPM Auth OTP? (Check 1PW) ');
+
+  await bumpVersion({ newVersion });
+  await build(); // run after the bumpVersion to have the latest version.
```
sandrina-p, line 105: note: Check my latest commit. https://github.com/remoteoss/json-schema-form/pull/3/commits/28792ddde1d9b12a50c36829f96e9996c941f079 I noticed a "missing feature" where the actual package would be published empty. Why? Because it was missing running the `build` script to generate the `dist/` during the release script.

sandrina-p: How you can test this (in this repo safely):

- run `npm run release:local` - and unzip the generated tarball. (this is explained in the read the CONTRIBUTING too)

To test the `npm run release:dev:*` or `npm run release:main`

- Comment the line with `await checkGitBranchAndStatus()` to bypass git checks
- Comment the line after the `await build` to ensure it doesn't try to commit or publish.

brennj: on it now

sandrina-p: Forgot to say, but I also created the `relese:local` script right before that commit, here: https://github.com/remoteoss/json-schema-form/pull/3/commits/fbadf2ea3039c3fcc0075dad1483a79cd08d360b

brennj: - `npm run release:local` works lovely and i've tested it in dragon and it shows up all in its glory.
- with the scary code behind comments, the `release:*:*` also are there... 

TBH for me this is good, 🚢 ?

brennj (approved):

scripts/release_dev.js
sandrina-p: Outcome from our Slack huddle:
- The first [dev release](https://www.npmjs.com/package/@remoteoss/json-schema-form) was a success!
- We open an MR in our internal codebase to replace the local imports with the released package.
- We noticed a couple of missing exports : (https://github.com/remoteoss/json-schema-form/pull/3/commits/8116f6fb7947e7c31d770681a5e5c488f8561bd5 and https://github.com/remoteoss/json-schema-form/pull/3/commits/8775cbf454a161de41ebd25b0f237a36e23af76f)

Tomorrow we will:
- Finish replacing the imports
- Wait for a green E2E in our internal codebase
- Merge this branch
- Release beta version in `main` :excited:

scripts/release.helpers.js
```diff
@@ -0,0 +1,126 @@
...
+  if (normalizedAnswer === 'y' || normalizedAnswer === 'yes') {
+    console.log('Confirmed! Proceeding...');
+    return 'yes';
+  }
+  if (normalizedAnswer === 'n' || normalizedAnswer === 'no') {
+    console.log('Cancelled. Exiting...');
+    return 'no';
+  }
+  console.log('Invalid input. Please enter Y or n.');
+  return askForConfirmation();
```
tugayac, line 59: Need to pass in `question` here again.

scripts/release.helpers.js
```diff
@@ -0,0 +1,126 @@
...
+  rl.close();
+  return answer;
+}
+
+async function askForConfirmation(question) {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+  const answer = await rl.question(`${question} (Y/n)`);
```
tugayac, line 45: The uppercase "Y" here implies the default is "yes" (so if I press enter, I expect to confirm, without having to type "Y"). However, I get an error:
```
Invalid input. Please enter Y or n.
```

If you want to prevent accidental entries, we can change it to just `y/n`?

scripts/release_dev.js
```diff
@@ -0,0 +1,111 @@
...
+
+  console.log('Creating a new dev...');
+  const versionType = process.argv.slice(2)[0]; // major | minor | patch
+  if (!versionType) {
+    console.log('🟠 version type is missing. Make sure to run the script from package.json');
+    process.exit(1);
+  }
+  const versionBase = semver.coerce(currentVersion); // 1.0.0-xxxx -> 1.0.0
+  return semver.inc(versionBase, versionType) + `-dev.${hour}`;
+}
```
tugayac, line 50: In case you were wondering why the commit SHA in the dev release version wasn't matching the current commit at the time the script is executed, this is why: https://www.loom.com/share/d5a02b42b04b407195e2b30311ba32c1

scripts/pr_dev_version.js
```diff
@@ -0,0 +1,32 @@
...
+  // TODO ideally I wanted to check diffs against "main",
+  // but I don't know how to do it.
+  if (version.includes('-dev')) {
+    console.log(
+      '🟠 This PR cannot be merged because the package.json version contains ".dev-". ' +
+        '\n   Please change the version back to the original one. ' +
+        '\n   Make sure the version is exactly the same as in the main branch.'
+    );
+    process.exit(1);
+  }
```
tugayac, line 16: Depending on what you want to check in terms of diffs, you might be able to use `git diff` with some flags to achieve what you want.

For example, you can do:
```
git diff --name-status --merge-base main HEAD
```
which will show what happened to the file + the file name, something like this:
```
M       mix.exs
D       priv/json_schemas/common/contract_details/order_all.json
M       priv/json_schemas/contract_details/AUS.json
A       priv/json_schemas/contract_details/AUT.json
```
It essentially shows what changed between the current HEAD and the commit that's the common ancestor of the current branch and `main` (assuming there's only a single common ancestor, which should be the case for us)

scripts/release_dev.js
sandrina-p: Nice TIL but you missed the point I want the SHA of the next commit (not existing yet). For example:

```bash
- 19ea0c1 "Release v0.1.0-dev.19ea0c1"
- 575f1a0 "feat: add potatoes"
- 9e21c3 "feat: add potatoes"
````

I need to know "19ea0c1" before I commit it so it matches in the commit msg.

But either way, even  if we find the answer, i'd rather stick with the date hash (yyyy..ss) because it's chronological and it was already useful yesterday to verify some versions.

scripts/pr_dev_version.js
sandrina-p: Today with fresh eyes I found a (different) way! Instead of using `git diff` to get the difference, I used `git show` to output the file content of main, and then parse it and manually compare it to the current branch. Check it out at https://github.com/remoteoss/json-schema-form/pull/3/commits/20b69f27eb14dcba86698766315a9538f9c8ae16

scripts/release.helpers.js
sandrina-p: Good point, fixed at https://github.com/remoteoss/json-schema-form/pull/3/commits/6019938c955c16636ea02e15a01640572db02563

scripts/release.helpers.js
sandrina-p: Excellent catch, fixed https://github.com/remoteoss/json-schema-form/pull/3/commits/6019938c955c16636ea02e15a01640572db02563

scripts/pr_dev_version.js
sandrina-p: Ah, although this worked locally, it [fails in CI](https://github.com/remoteoss/json-schema-form/actions/runs/5012270770/jobs/8984036926?pr=3) because it can't perform `git` commands. So I reverted the `pr_dev_version.js` to what I had before https://github.com/remoteoss/json-schema-form/pull/3/commits/241dfc2a37c966cbb7b5022f65b08ed2b4820716.

sandrina-p: Okay, now is working file, with [CI throwing the expected error](https://github.com/remoteoss/json-schema-form/actions/runs/5012509498/jobs/8984543336?pr=3). I've pushed a last commit to revert it back to `0.0.0` using `npm run version_as_main`. https://github.com/remoteoss/json-schema-form/pull/3/commits/241dfc2a37c966cbb7b5022f65b08ed2b4820716

scripts/release_dev.js
tugayac: Ah got it! Yeah I like the date-based version, that's a good idea 💡

PR #6, opened 2023-05-18
feat: added json-schema-form meta schema

jguddas:
Pretty nice stuff you made here.

I was missing some type annotations for those custom `x-jsf-` properties, so I wrote some.

I hope there isn't already a definition I overlooked?

Screenshot of editor showing an error message when hovering x-jsf-presentation.inputType due to it being set to an unsupported type, in this case textarea.

### DIFF
--- a/json-schema-form.schema.json
+++ b/json-schema-form.schema.json
@@ -96,7 +96,7 @@
         "inputType": {
           "description": "Input type for the generated UI field",
           "type": "string",
-          "oneOf": [
+          "anyOf": [
             {
               "const": "text",
               "description": "Similar to the native HTML input with text type."
--- a/json-schema-form.schema.json
+++ b/json-schema-form.schema.json
@@ -92,10 +92,9 @@
     "x-jsf-presentation": {
       "type": "object",
       "description": "Presentation overrides for the schema",
-      "additionalProperties": false,
       "properties": {
         "inputType": {
-          "description": "Input type for the schema",
+          "description": "Input type for the generated UI field",
           "type": "string",
           "oneOf": [
             {
@@ -137,6 +136,10 @@
             {
               "const": "group-array",
               "description": "A list of inputs that can be repeated. Its expected value is an array."
+            },
+            {
+              "type": "string",
+              "description": "Any arbitrary custom inputType you might want to represent your UI Field."
             }
           ]
         },
--- a/json-schema-form.schema.json
+++ b/json-schema-form.schema.json
@@ -1,0 +1,153 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "JSON json-schema-form-Schema",
+  "definitions": {
+    "schemaArray": {
+      "allOf": [
+        {
+          "$ref": "http://json-schema.org/draft-07/schema#/definitions/schemaArray"
+        },
+        {
+          "items": { "$ref": "#" }
+        }
+      ]
+    }
+  },
+  "allOf": [{ "$ref": "http://json-schema.org/draft-07/schema#" }],
+  "properties": {
+    "additionalItems": { "$ref": "#" },
+    "additionalProperties": { "$ref": "#" },
+    "dependencies": {
+      "additionalProperties": {
+        "anyOf": [{ "$ref": "#" }, { "type": "array" }]
+      }
+    },
+    "items": {
+      "anyOf": [{ "$ref": "#" }, { "$ref": "#/definitions/schemaArray" }]
+    },
+    "definitions": {
+      "additionalProperties": { "$ref": "#" }
+    },
+    "patternProperties": {
+      "additionalProperties": { "$ref": "#" }
+    },
+    "properties": {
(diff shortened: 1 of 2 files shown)


json-schema-form.schema.json
```diff
@@ -0,0 +1,153 @@
...
+              "description": "Similar to the native HTML input with file type."
+            },
+            {
+              "const": "fieldset",
+              "description": "Groups multiple Fields inside. Its expected value is a nested object."
+            },
+            {
+              "const": "group-array",
+              "description": "A list of inputs that can be repeated. Its expected value is an array."
+            }
```
sandrina-p, line 112: todo: Following the example above with the "money field". If the dev wants to represent a money field, probably it would look like this:

```jsonc
{
  "properties": {
    "budget": {
      "title": "What's your budget?",
      "type": "number",
      "x-jsf-presentation": {
        "inputType": "money", // <-- a custom inputType!
        "currency": "EUR"
      }
    }
  }
}
```

This means that, although JSF tries to guess the best field's `inputType` based on each property schema (when `x-jsf-presenttion` is not declared), it also allows you to pass any arbitrary value you need in your UI Form.

In other words, your meta schema is missing a last `oneOf`:

```suggestion
            },
            {
              "type": "string",
              "description": "Any arbitrary custom inputType you might want to represent your UI Field."
            }
```

The `inputType` does not influence JSF internals in any way. It's just for a UI flavor. Gotcha? ☺️

sandrina-p (changes requested):
Hi @jguddas, nice to reconnect again, this time about JSON Schemas! 🙌 

Thanks a lot for this meta schema, is pretty complete! Left a few notes with a couple examples. Let me know if it makes sense :)

json-schema-form.schema.json
```diff
@@ -0,0 +1,153 @@
...
+    "then": { "$ref": "#" },
+    "else": { "$ref": "#" },
+    "allOf": { "$ref": "#/definitions/schemaArray" },
+    "anyOf": { "$ref": "#/definitions/schemaArray" },
+    "oneOf": { "$ref": "#/definitions/schemaArray" },
+    "not": { "$ref": "#" },
+    "contains": { "$ref": "#" },
+    "propertyNames": { "$ref": "#" },
+    "x-jsf-order": {
+      "description": "This keyword defines the order of fields for a given form or fieldset. It's placed at the schema root and inside of each fieldset.",
```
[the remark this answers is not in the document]
jguddas: First of all, hi again :wave:!

I can inline the definition, but for now it already works due to this ref here.
```
"$ref": "http://json-schema.org/draft-07/schema#/definitions/stringArray"
```

json-schema-form.schema.json
jguddas: Does that mean I can also customize things like `statement` and `statement.severity`?

And is there anything that we should mark as required?

jguddas (commented):
If you are happy with it, we can also think about submitting it to [SchemaStore](https://github.com/SchemaStore/schemastore/tree/master/src/schemas/json).

It might be the better place to store it over having it in the root of this project.

json-schema-form.schema.json
sandrina-p: > Does that mean I can also customize things like statement and statement.severity?

Yes, exactly. We could say that `statement` is our ([Remote](https://remote.com/)) opinionated [(re)presentation of statements](https://json-schema-form.vercel.app/?path=/story/demos-fields--field-with-statement) in a form. In a future version, we could find a way to [remove it from JSF internals](https://github.com/remoteoss/json-schema-form/blob/main/src/helpers.js#LL455C9-L455C29)

> And is there anything that we should mark as required?

Good question. The only required key is `properties` in the root, as it's supposed to be a form (list) of fields.

JSF works totally fine with the most basic shape for each json schema property. (you can test it on the [playground](https://json-schema-form.vercel.app/?path=/docs/playground--playground)). Here's an example:

```json
{
  "properties": {
    "username": {},
    "age": {}
  }
}
```

json-schema-form.schema.json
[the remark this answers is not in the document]
jguddas: It's useful for when there is some form of standard, for example all files inside `.github/workflows` automatically get the schema from [schemas/json/github-workflow.json](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-workflow.json) in modern editors.

But I don't think it's super necessary, the validation and intellisense also work with it just being set as `$schema`.

https://github.com/remoteoss/json-schema-form/assets/25524993/ff886dfd-55ed-498f-a9bc-460f2df81c88

json-schema-form.schema.json
```diff
@@ -0,0 +1,188 @@
+{
```
sandrina-p, line 1: Thanks a lot, @jguddas for your contribution! You were the first one 😄 Approved, merging it now :)

jguddas: I think you are right, meta schema is the right term.

sandrina-p (approved):

PR #10, opened 2023-06-04
fix: support maximum 0 in number field

jguddas:


src/tests/createHeadlessForm.test.js
```diff
@@ -780,6 +781,29 @@ describe('createHeadlessForm', () => {
       expect(() => fieldValidator.validateSync('')).toThrowError('The value must be a number');
     });
 
+    it('shows an error when positive number is entered into a "number" field with maximum set to zero', () => {
```
sandrina-p, line 784: suggestion: First of all, we know that this file deserves some love as it's huge and a little messy. But could I ask to move this test to inside the `describe('when a field is number'` (line ~1812) so it's closer to the other ones related to min/max assertions?

Regarding validating the field itself, the way you did is the original way we did it, but now we use `handleValidation` directly to be closer to what the tool actually does. You can check "'applies correct validation for fieldset fields'" as an example. For you case, it would look like this:

```
const { handleValidation } = createHeadlessForm(schemaInputTypeNumberZeroMaximum);
const validateForm = (vals) => friendlyError(handleValidation(vals));

expect(validateForm({ tabs: 1 })).toEqual({
  tabs: 'Must be smaller or equal to 0',
});

```

src/tests/helpers.js
```diff
@@ -46,6 +46,23 @@ export const mockNumberInput = {
...
+  description: 'How many open tabs do you have?',
+  'x-jsf-presentation': {
+    inputType: 'number',
+  },
+  minimum: -100,
+  maximum: 0,
+  type: 'number',
+};
+
+export const schemaInputTypeNumberZeroMaximum = JSONSchemaBuilder()
```
sandrina-p, line 60: suggestion: `JSONSchemaBuilder` was something we created earlier in the development, but we've been removing it as it creates an extra layer of complexity. In this case, it could be simply:

```js
export const schemaInputTypeNumberZeroMaximum = {
  properties: {
    tabs: {
      title: 'Tabs',
      description: 'How many open tabs do you have?',
      'x-jsf-presentation': {
        inputType: 'number',
      },
      minimum: -100,
      maximum: 0,
      type: 'number',
    },
  },
};
```

P.S. We don't even need to import `mockNumberInputNegative` as it's not used anywhere else.

sandrina-p (changes requested):
Thanks for this PR! Left just a couple of comments :)

src/tests/createHeadlessForm.test.js
```diff
@@ -780,6 +781,29 @@ describe('createHeadlessForm', () => {
...
+      const result = createHeadlessForm(schemaInputTypeNumberZeroMaximum);
+      expect(result).toMatchObject({
+        fields: [
+          {
+            description: 'How many open tabs do you have?',
+            label: 'Tabs',
+            name: 'tabs',
+            required: false,
+            schema: expect.any(Object),
+            type: 'number',
```
sandrina-p, line 794: todo: `type` is deprecated. Instead, can you assert `jsonType` and `inputType`?

src/tests/createHeadlessForm.test.js
```diff
@@ -1839,6 +1863,16 @@ describe('createHeadlessForm', () => {
         ).resolves.toEqual(assertObj);
       });
     });
+    describe('and maximum is set to zero', () => {
```
sandrina-p, line 1866: question: This new test is great, do you plan to keep the one added in previous commit too? I'd say just one merging all assertions is better. what do you think?

jguddas: Woops, I just forgot to get rid of the old test cases.

Do you have a ```suggestion on what other cases I could add here?

sandrina-p: I'd sy that the cases you wrote in the original test (0, 10, -11 5) are perfect. Just move them here and rename the description to `it("shows the correct validation")`

PR #14, opened 2023-06-16
feat: Add Typescript declarations

gabrielseco:
## 1. Description
Add a declaration file giving some types to the 3 exported functions that the library has.

* `createHeadlessForm`
* `pickXKey`
* `buildCompleteYupSchema`

Some arguments are generic to be able to ship something fast, I think if we need better types, we'll need to spend more time on it.

## 2. How to test this

I created a next app with typescript in my local environment and copied the content of this [codesandbox](https://codesandbox.io/s/json-schema-form-demo-react-igh3sk?from-embed), after doing that, I installed a release dev version, the one I used is 0.1.1-dev.20230616085129 to test that everything work as expected.

### DIFF
--- a/json-schema-form.d.ts
+++ b/json-schema-form.d.ts
@@ -16,7 +16,7 @@ type JSConfig = {
   }
 }
 
-type Fields = any; // TODO: We don't know what type we have here, we need to investigate
+type Fields = Record<string, unknown>[]; // TODO: We don't know what type we have here, we need to investigate
 
 type $TsFixMe = any;
 
@@ -25,5 +25,4 @@ type $TsFixMe = any;
  * These fields must be the same from
  * const { fields } = createHeadlessForm()
  */
-
 export function buildCompleteYupSchema(fields: Fields, config: JSConfig): $TsFixMe // TODO: We don't what yup returns here, we'll fix it later
--- a/json-schema-form.d.ts
+++ b/json-schema-form.d.ts
@@ -3,18 +3,50 @@
  */
 export function pickXKey(node: Object, key: 'presentation' | 'errorMessage'): Object | undefined;
 
+type ValidationTypes =
+  | 'type'
+  | 'minimum'
+  | 'maximum'
+  | 'minLength'
+  | 'maxLength'
+  | 'pattern'
+  | 'maxFileSize'
+  | 'accept'
+  | 'required';
+
 type JSFConfig = {
+  /**
+   * Initial json values to prefill the form fields.
+   * This influences the initial visibility of conditional fields.
+   */
   initialValues?: Record<string, unknown>;
+  /**
+   * It inforces all json properties (fields) to have x-jsf-presentation.inputType.
+    @default true 
+  */
   strictInputType?: boolean;
+  /**
+   * Customize the output of a given named Field.
+   * Useful if you want to enhance the UI/UX of a particular field.
+   * @example
+   * { has_pet: { "optionsDirection": "horizontal" } }
+   */
   customProperties?: {
     description: (description: string, field: $TSFixMe) => string | string;
     [key: string]: unknown;
   };
+  /**
+   * Customize the config for each inputType
+   */
   inputTypes?: {
-    errorMessage: Record<string, unknown>;
+    /**
+     * Customize the error error message of this input type.
+     * @example
+     * { required: "This number is required." }
+     */
+    errorMessage?: Record<ValidationTypes, string>;
   };
 };
-
 type Fields = Record<string, unknown>[]; //TODO: Type the field based on the given JSON Schema properties.
 
 type $TSFixMe = any;
@@ -27,7 +59,42 @@ type $TSFixMe = any;
 export function buildCompleteYupSchema(fields: Fields, config: JSFConfig): $TSFixMe; //TODO: We need to update Yup to 1.0 which supports TS.
 
 type HeadlessFormOutput = {
+  /**
+   * List of Fields. Each Field corresponds to a form input from the json schema.
+   * @example
+   * [
+      {
+        name: "has_pet",
+        label: "Has Pet",
+        options: [
+          { label: "Yes", value: "yes" },
+          { label: "No", value: "no" }
+        ],
+        required: true,
+        inputType: "radio",
+        jsonType: "string",
+        description: "Do you have a pet?",
+        errorMessage: {},
+        isVisible: true
+      },
+      {
+        type: "text",
+        label: "Pet's name",
+        description: "What's your pet's name?",
+        required: false,
+        inputType: "text",
+        jsonType: "string",
+        isVisible: false
+      }
+   * ]
+   */
   fields: Fields;
+  /**
+   * Validates given values and returns the respective errors.
+   * @example
+   * const { formErrors } = handleValidation({ has_pet: "yes" });
+   * console.log(formErrors) // { pet_name: "Required field." }
+   */
   handleValidation: () => $TSFixMe;
   isError: boolean;
   error?: Error;
--- a/json-schema-form.d.ts
+++ b/json-schema-form.d.ts
@@ -1,4 +1,4 @@
 /**
  * Shorthand to lookup for keys with `x-jsf-*` preffix.
  */
-export function pickXKey(node: Object, key: 'presentation' | 'error-message'): Object
+export function pickXKey(node: Object, key: 'presentation' | 'error-message'): Object | undefined
--- a/json-schema-form.d.ts
+++ b/json-schema-form.d.ts
@@ -7,7 +7,7 @@ type JSConfig = {
   initialValues?: Record<string, unknown>;
   strictInputType?: boolean;
   customProperties?: {
-    description: Function | string;
+    description: (description: string, field: $TSFixMe) => string | string;
     [key: string]: unknown;
   };
   inputTypes?: {
--- a/json-schema-form.d.ts
+++ b/json-schema-form.d.ts
@@ -2,3 +2,28 @@
  * Shorthand to lookup for keys with `x-jsf-*` preffix.
  */
 export function pickXKey(node: Object, key: 'presentation' | 'error-message'): Object | undefined
+
+
+type JSConfig = {
+  initialValues?: Record<string, unknown>;
+  strictInputType?: boolean;
+  customProperties?: {
+    description: () => void | string;
+    [key: string]: unknown;
+  }
+  inputTypes?: {
+    errorMessage: Record<string, unknown>;
+  }
+}
+
+type Fields = any; // TODO: We don't know what type we have here, we need to investigate
+
+type $TsFixMe = any;
+
+/**
+ * Returns the Yup schema structure of given fields.
+ * These fields must be the same from
+ * const { fields } = createHeadlessForm()
+ */
+
+export function buildCompleteYupSchema(fields: Fields, config: JSConfig): $TsFixMe // TODO: We don't what yup returns here, we'll fix it later
--- a/json-schema-form.d.ts
+++ b/json-schema-form.d.ts
@@ -26,23 +26,17 @@ type $TsFixMe = any;
  */
 export function buildCompleteYupSchema(fields: Fields, config: JSConfig): $TsFixMe; // TODO: We don't what yup returns here, we'll fix it later
 
-type HeadlessFormOutput =
-  | {
-      fields: Fields;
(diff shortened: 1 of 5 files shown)


json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Shorthand to lookup for keys with `x-jsf-*` preffix.
+ */
+export function pickXKey(node: Object, key: 'presentation' | 'error-message'): Object | undefined;
```
sandrina-p, line 4: todo: It's `errorMessage`, not `error-message`

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type JSConfig = {
+  initialValues?: Record<string, unknown>;
+  strictInputType?: boolean;
+  customProperties?: {
+    description: Function | string;
```
sandrina-p, line 10: question: Can we type this Function better? It must return a string. The Fn params are `(description: string, field: object)`. The `field` is complex to type, we could use the `$TsFixMe`
([Source code](https://github.com/remoteoss/json-schema-form/blob/main/src/internals/helpers.js#L42))

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    [key: string]: unknown;
+  };
+  inputTypes?: {
+    errorMessage: Record<string, unknown>;
+  };
+};
+
+type Fields = Record<string, unknown>[]; // TODO: We don't know what type we have here, we need to investigate
+
+type $TsFixMe = any;
```
sandrina-p, line 20: nit: Could we name it `$TSFixMe` to match our internal codebase?

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Returns the Yup schema structure of given fields.
+ * These fields must be the same from
+ * const { fields } = createHeadlessForm()
+ */
+export function buildCompleteYupSchema(fields: Fields, config: JSConfig): $TsFixMe; // TODO: We don't what yup returns here, we'll fix it later
```
sandrina-p, line 27: extra note:

```suggestion
export function buildCompleteYupSchema(fields: Fields, config: JSConfig): $TsFixMe; // TODO: We need to update Yup to 1.0 which supports TS.
```

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type HeadlessFormOutput = {
+  fields: Fields;
+  handleValidation: () => $TsFixMe;
+  isError: boolean;
+  error?: undefined;
```
sandrina-p, line 33: todo: The error is a `Error`

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+  customProperties?: {
+    description: Function | string;
+    [key: string]: unknown;
+  };
+  inputTypes?: {
+    errorMessage: Record<string, unknown>;
+  };
+};
+
+type Fields = Record<string, unknown>[]; // TODO: We don't know what type we have here, we need to investigate
```
sandrina-p, line 18:
```suggestion
type Fields = Record<string, unknown>[]; // TODO: Type the field based on the given JSON Schema properties.
```

Overall, let's not say "we don't know" because we do... (it's documented in JSDocs). We just don't have the time to type it properly right now 😅

sandrina-p (changes requested):

json-schema-form.d.ts
gabrielseco: Thanks, I didn't know what version started to support typescript

json-schema-form.d.ts
gabrielseco: Done!

json-schema-form.d.ts
gabrielseco: Done!

json-schema-form.d.ts
gabrielseco: Done!

json-schema-form.d.ts
gabrielseco: In the JSDoc, I saw the fields property try to use a type called Fields but it pointed to any, that's why I wrote we needed to investigate more.

Changed

json-schema-form.d.ts
gabrielseco: Done!

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Shorthand to lookup for keys with `x-jsf-*` preffix.
+ */
+export function pickXKey(node: Object, key: 'presentation' | 'errorMessage'): Object | undefined;
+
+type JSConfig = {
```
johnstonbl01, line 6: **nit:** Could we rename this type to be something a bit more descriptive? When I first read this, my mind immediately thought "JavaScript" and not "JsonSchema." Maybe `CustomFormConfig` or `CustomJSONConfig`. Thoughts @sandrina-p ?

johnstonbl01 (commented):
Just a small nitpick after addressing Sandrina's comments. Nice work! 👍

json-schema-form.d.ts
gabrielseco: I missed an F there,

In the JSDOC this is called JSFConfig, I don't mind changing the name though I think Config would be enough

json-schema-form.d.ts
sandrina-p: Ah, I was wrong, afterall it's after [v0.32](https://github.com/jquense/yup/blob/master/CHANGELOG.md#0320-2020-12-03)

json-schema-form.d.ts
sandrina-p: Yes, let's stick with `JSFConfig`, as it's what we use in multiple files and written docs :)

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+  handleValidation: () => $TSFixMe;
+  isError: boolean;
+  error?: Error;
+};
+
+/**
+ * Generates the Headless form based on the provided JSON schema
+ */
+export function createHeadlessForm(
+  jsonSchema: Record<string, unknown>,
```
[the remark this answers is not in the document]
gabrielseco: Done!

json-schema-form.d.ts
```diff
@@ -0,0 +1,42 @@
...
+  customConfig: JSFConfig
```
sandrina-p, line 43: bug: `customConfig` is optional!

json-schema-form.d.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Shorthand to lookup for keys with `x-jsf-*` preffix.
+ */
+export function pickXKey(node: Object, key: 'presentation' | 'errorMessage'): Object | undefined;
+
+type JSFConfig = {
+  initialValues?: Record<string, unknown>;
```
sandrina-p, line 7: docs: They are missing, I've added them. Also, the `errorMessage` type can be improved. All is here:

```js

type ValidationTypes = 'type' | 'minimum' | 'maximum' | 'minLength' | 'maxLength' | 'pattern' | 'maxFileSize' | 'accept';

type JSFConfig = {
  /**
   * Initial json values to prefill the form fields.
   * This influences the initial visibility of conditional fields.
  */
  initialValues?: Record<string, unknown>;
  /**
   * It inforces all json properties (fields) to have x-jsf-presentation.inputType.
    @default true 
  */
  strictInputType?: boolean;
  /**
   * Customize the output of a given named Field.
   * Useful if you want to enhance the UI/UX of a particular field.
   * @example
   * { has_pet: { "optionsDirection": "horizontal" } }
  */
  customProperties?: {
    description: (description: string, field: $TSFixMe) => string | string;
    [key: string]: unknown;
  };
  /**
   * Customize the config for each inputType
  */
 inputTypes?: {
    /**
     * Customize the error error message of this input type.
     * @example
     * { required: "This number is required." }
     */
    errorMessage?: Record<ValidationTypes, string>;
  };
};
```

sandrina-p: P.S. The `Record<ValidationTypes, string>` seems incorrect, what I mean is that each key must one of one the ValidationTypes. #LearningTS

json-schema-form.d.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Returns the Yup schema structure of given fields.
+ * These fields must be the same from
+ * const { fields } = createHeadlessForm()
+ */
+export function buildCompleteYupSchema(fields: Fields, config: JSFConfig): $TSFixMe; //TODO: We need to update Yup to 1.0 which supports TS.
+
+type HeadlessFormOutput = {
+  fields: Fields;
+  handleValidation: () => $TSFixMe;
```
sandrina-p, line 31: docs: More docs/example for these ones:

<details>

```js
type HeadlessFormOutput = {
  /**
   * List of Fields. Each Field corresponds to a form input from the json schema.
   * @example
   * [
      {
        name: "has_pet",
        label: "Has Pet",
        options: [
          { label: "Yes", value: "yes" },
          { label: "No", value: "no" }
        ],
        required: true,
        inputType: "radio",
        jsonType: "string",
        description: "Do you have a pet?",
        errorMessage: {},
        isVisible: true
      },
      {
        type: "text",
        label: "Pet's name",
        description: "What's your pet's name?",
        required: false,
        inputType: "text",
        jsonType: "string",
        isVisible: false
      }
   * ]
   */
  fields: Fields;
  /**
   * Validates given values and returns the respective errors.
   * 
   * @example
   * const { formErrors } = handleValidation({ has_pet: "yes" });
   * console.log(formErrors) // { pet_name: "Required field." } 
  */
  handleValidation: () => $TSFixMe;
  isError: boolean;
  error?: Error;
};
```

</details>

json-schema-form.d.ts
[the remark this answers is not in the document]
sandrina-p: A small nit: Add written docs too   /** A JSON Schema of type object */

json-schema-form.d.ts
gabrielseco: I have a couple of doubts here:

* if we give `inputTypes.errorMessage`  the value of` { required: "This number is required." }`, it'll fail as required is not present in the `ValidationTypes`.

* After reviewing the JSFConfig type, should the description property inside customProperties be optional instead of required?

sandrina-p (changes requested):
Okay, review done, I tested the latest dev release with the `.ds` file replaced. Most errors were about keys that were optional. I've also suggested some docs/examples! Thanks Gabriel!

json-schema-form.d.ts
sandrina-p: - I forgot to add `required` to `ValidationTypes` 🤦 
- You are right, `description` is optional too yes! Basically every key of `JSFConfig` is optional.

gabrielseco: Fixed it

sandrina-p: Hum, I believe something is still wrong because this is valid but TS marks it as invalid:

```js
const {fields, handleValidation} = createHeadlessForm(demo, {
    inputTypes: {
      errorMessage: {
        required: 'sss',
      },
    },
  });
```

image

json-schema-form.d.ts
sandrina-p: @gabrielseco maybe you missed this comment?

json-schema-form.d.ts
gabrielseco: Done!

json-schema-form.d.ts
gabrielseco: We were telling typescript that we needed an object with all the `ValidationTypes` keys.

Fixed that with `Partial`

json-schema-form.d.ts
gabrielseco: Done!

json-schema-form.d.ts
sandrina-p: TIL about `Partial`. it's perfect now!

sandrina-p (changes requested):
Ready to go! 

Edit: Ups, pressed the wrong button.

sandrina-p (approved):
Ready go to!

json-schema-form.d.ts
```diff
@@ -0,0 +1,115 @@
+/**
```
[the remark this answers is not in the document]
gabrielseco: It just came the notification after I squashed

PR #15, opened 2023-06-17
WIP: fix: crashing when format is uri

jguddas:
`{
            type: 'string',
            format: 'uri'
          }` results in the following error:
```
  ● createHeadlessForm › when a field is text › and format is uri › should not throw an error

    expect(jest.fn()).not.toHaveBeenCalled()

    Expected number of calls: 0
    Received number of calls: 1

    1: "JSON Schema invalid!", [TypeError: Cannot convert undefined or null to object]

      82 |
      83 | afterEach(() => {
    > 84 |   expect(console.error).not.toHaveBeenCalled();
         |                             ^
      85 |   console.error.mockRestore();
      86 | });
      87 |

      at Object.toHaveBeenCalled (src/tests/createHeadlessForm.test.js:84:29)

```

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -1810,6 +1810,22 @@ describe('createHeadlessForm', () => {
     });
   });
 
+  describe('when a field is text', () => {
+    describe('and format is uri', () => {
+      it('should not throw an error', () => {
+        const { handleValidation } = createHeadlessForm({
+          url: {
+            type: 'string',
+            format: 'uri',
+          },
+        });
+        const validateForm = (vals) => friendlyError(handleValidation(vals));
+
+        expect(validateForm({ url: 'https://example.com' })).toBeUndefined();
+      });
+    });
+  });
+
   describe('when a field is number', () => {
     let fields;
     beforeEach(() => {


src/tests/createHeadlessForm.test.js
```diff
@@ -1810,6 +1810,22 @@ describe('createHeadlessForm', () => {
     });
   });
 
+  describe('when a field is text', () => {
```
sandrina-p, line 1813: Hi Jakob! I'm not understanding, is this a bug report or a PR with the actual fix that you forgot to push? 😅

jguddas: I can open an issue for it, it's just a small bug, but I don't have much time to fix it right now, just pushed the WIP state for now.

It will probably stay as draft for a couple more weeks, you should have access to `jguddas:fix/crashing-when-format-is-uri` if you want to fix this quick.

sandrina-p: I don't have time either, it's okay to stay here for a couple of months ;)

sandrina-p:
We are closing the PR, feel free to reopen it when you have a proper solution :)

PR #16, opened 2023-06-17
WIP: feat: support nested group arrays

jguddas:
![storybook playground showing steps[0].undefined as a field instead of a nested array](https://github.com/remoteoss/json-schema-form/assets/25524993/d0228dea-8bd9-4151-9bd8-af2871168ea9)

`buildGroupArraySchema` always expects an object, but it could be something else.
https://github.com/remoteoss/json-schema-form/blob/1e9d6cf96436cf16018e045b351567c643e10dac/src/yupSchema.js#L207-L217

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -1008,6 +1008,48 @@ describe('createHeadlessForm', () => {
         ]);
       });
 
+      it('supports nested group arrays', () => {
+        const result = createHeadlessForm({
+          type: 'object',
+          required: ['steps'],
+          properties: {
+            steps: {
+              type: 'array',
+              required: ['ingredients'],
+              additionalItems: false,
+              'x-jsf-presentation': {
+                inputType: 'group-array',
+                addFieldText: 'Add new step',
+              },
+              items: {
+                type: 'object',
+                additionalProperties: true,
+                properties: {
+                  highlights: {
+                    type: 'array',
+                    additionalItems: false,
+                    'x-jsf-presentation': {
+                      inputType: 'group-array',
+                      addFieldText: 'Add new ingredient',
+                    },
+                    items: {
+                      type: 'string',
+                      'x-jsf-presentation': {
+                        inputType: 'string',
+                      },
+                    },
+                  },
+                },
+              },
+            },
+          },
+        });
+
+        expect(result.fields[0].schema.innerType.fields).not.toMatchObject({
+          undefined: undefined,
+        });
+      });
+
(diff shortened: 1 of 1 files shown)


jguddas:
Something like this is also currently broken `{ steps: ["a", "b", "c"] }`.

```json
{
  "type": "object",
  "required": ["steps"],
  "properties": {
    "steps": {
      "type": "array",
      "x-jsf-presentation": {
        "inputType": "group-array",
        "addFieldText": "Add new step"
      },
      "items": {
        "type": "string"
      }
    }
  }
}
```

src/tests/createHeadlessForm.test.js
```diff
@@ -1008,6 +1008,48 @@ describe('createHeadlessForm', () => {
...
+          properties: {
+            steps: {
+              type: 'array',
+              required: ['ingredients'],
+              additionalItems: false,
+              'x-jsf-presentation': {
+                inputType: 'group-array',
+                addFieldText: 'Add new step',
+              },
+              items: {
```
sandrina-p, line 1024: it's true Jakob, this is something that we didn't build because we don't have nested group-arrays at Remote.

jguddas: It's not just groups, it's any kind of array with children that are not objects, so `string[]` or `number[]` don't work either.

sandrina-p: I see. Visually speaking, every form field must be an `object` to declare at least the `title` (label), that's why it's made to support just `object[]`. In the same way that the `jsonSchema` passed to `createHeadlessForm()` must also be an object with `properties`, otherwise it will throw.

jguddas:
```
{
  "type": "object",
  "required": ["steps"],
  "title" : "Steps to reproduce the bug",
  "description": "What did you do to cause this?",
  "properties": {
    "steps": {
      "type": "array",
      "x-jsf-presentation": {
        "inputType": "group-array",
        "addFieldText": "Add new step"
      },
      "items": {
        "title": "Step",
        "type": "string"
      }
    }
  }
}
```

## Could result in:

---

### Steps to reproduce the bug

What did you do to cause this?

1. Step
  `Open app`
2. Step
  `Go to foo page`
3. Step
  `Press bar button|`

<kbd>+ Add new step</kbd>

<kbd>Submit</kbd>

sandrina-p:
Hi @jguddas. We are closing the PR because it's only reporting a bug but not actually fixing it. Feel free to open an issue with the same bug report, or reopen a PR with a working solution :)

PR #17, opened 2023-06-20
feat(Radio/Select): Spread x-jsf-presentation value to option root

sandrina-p:
### Example 
For example, given the following JSON Schema, where each `oneOf` contains a `x-jsf-presentation`:

```js
{
  properties: {
    health_perks: {
      title: 'Health perks',
      oneOf: [
        {
          const: 'basic',
          title: 'Basic',
          'x-jsf-presentation': {
            cost: '$30.00/mo',
          },
          'x-another': 'extra-thing',
        },
        {
          const: 'standard',
          title: 'Standard',
          'x-jsf-presentation': {
            cost: '$50.00/mo',
          },
        }
      ],
      'x-jsf-presentation': {
        inputType: 'radio',
        layout: "vertical"
      },
      type: 'string',
    },
  },
}
```

Without this PR, to access the `cost` of each option, you'd need to do:

``` js
const { fields } = createHeadlessForm(demoAbove);

const cost = fields[0].options[0]['x-jsf-presentation'].cost
```

Now, you can access it directly with:

```js
const cost = fields[0].options[0].displayCost
```

In other words, now we move the `x-jsf-presentation`'s data to the closest parent, removing the key itself, similar to how we already do with the root (check `layout` to compare). The `fields[0]` would look like this:

```js
{
  "name": "health_perks",
  "label": "Health perks",
  "options": [
    {
      "label": "Basic",
      "value": "basic",
      "cost": "$30.00/mo", // <-- moved to parent
      "x-another": "extra-thing" // <-- still accepts any other arbitrary key
    },
    {
      "label": "Standard",
      "value": "standard",
      "cost": "$50.00/mo"
    }
  ],
  "inputType": "radio",
  "jsonType": "string",
  "layout": "vertical"  // <-- the new behavior is similar to this existing one in the field root
}
```

---

## Related
- [Linear issue](https://linear.app/remote/issue/BNFT-670) (private)
- [Slack thread](https://remote-com.slack.com/archives/C02HTN0LY02/p1687270322322069) (private)

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -406,12 +406,27 @@ export function updateFieldsProperties(fields, formValues, jsonSchema) {
 
 const notNullOption = (opt) => opt.const !== null;
 
+function flatPresentation(item) {
+  return Object.entries(item).reduce((newItem, [key, value]) => {
+    if (key === 'x-jsf-presentation') {
+      return {
+        ...newItem,
+        ...value,
+      };
+    }
+    return {
+      ...newItem,
+      [key]: value,
+    };
+  }, {});
+}
+
 function getFieldOptions(node, presentation) {
   function convertToOptions(nodeOptions) {
     return nodeOptions.filter(notNullOption).map(({ title, const: cons, ...item }) => ({
       label: title,
       value: cons,
-      ...item,
+      ...flatPresentation(item),
     }));
   }
 
--- a/src/tests/helpers.js
+++ b/src/tests/helpers.js
@@ -912,9 +912,9 @@ export const schemaInputTypeRadioOptionsWithDetails = {
           const: 'basic',
           title: 'Basic',
           'x-jsf-presentation': {
-            carrierName: 'Segure Inc',
-            displayCost: '$30.00/mo',
-            urlDetails: 'www.example-bsc.com',
+            meta: {
+              displayCost: '$30.00/mo',
+            },
           },
           'x-another': 'extra-thing',
         },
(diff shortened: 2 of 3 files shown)


src/helpers.js
```diff
@@ -406,12 +406,27 @@ export function updateFieldsProperties(fields, formValues, jsonSchema) {
 
 const notNullOption = (opt) => opt.const !== null;
 
+function flatPresentation(item) {
```
sandrina-p, line 409: Ready for review :)

src/tests/helpers.js
```diff
@@ -871,34 +871,80 @@ export const schemaInputDeprecated = JSONSchemaBuilder()
...
+            displayCost: '$30.00/mo',
+            urlDetails: 'www.example-bsc.com',
+          },
+          'x-another': 'extra-thing',
+        },
+        {
+          const: 'standard',
+          title: 'Standard',
+          'x-jsf-presentation': {
+            carrierName: 'Vanilla Lda',
```
brennj, line 925: for consistency sake in regards to internal discussions, should we wrap this in `meta` just to have everything aligned in case someone combs through the docs?

brennj: (this is the only comment i have, i'm happy with this otherwise)

sandrina-p: Yes, done!

brennj (approved):
lets go!

PR #18, opened 2023-06-22
fix(select/radio): Accept just the values in options (plus `''` and `null` for backward-compatibility)

gabrielseco:
## 1. Description
Fix select and radio yup schemas to only validate the values provided by a json schema

## 2. How to test this

~~I created a playground with vite where you can test the different fields like select and radio but I think we should remove it from this PR and moved it to a new one to have a proper release with a good developer experience.~~

The playground demo is based on the [codesandbox](https://codesandbox.io/s/json-schema-form-demo-react-igh3sk?from-embed) from our docs

PS: I'll be on PTO from tomorrow, I'll be back on the 3th of July

Edit by Sandrina:
I've added more automated tests to cover all the scenarios.


src/yupSchema.js
```diff
@@ -34,8 +34,16 @@ const validateOnlyStrings = string()
 
 const yupSchemas = {
```
sandrina-p, line 35: This fix will be fundamental to go along with https://github.com/remoteoss/json-schema-form/pull/23 and ship in the next release for Remote's API Partners. I'm picking this PR to make it ready for review tomorrow. The plan:
1. Remove the playground to another branch/PR
2. Mark it as ready
3. At https://github.com/remoteoss/json-schema-form/pull/23, rebase with this one (branch of a branch) to improve the tests severity and deploy a dev-release with both together.

src/tests/helpers.js
```diff
@@ -127,28 +127,6 @@ export const mockRadioInput = {
   },
   type: 'string',
 };
-export const mockRadioInputOptional = {
```
sandrina-p, line 130: note: This wasn't deleted, it was just moved to line 866 for better readability.

src/yupSchema.js
```diff
@@ -74,9 +78,58 @@ function getRequiredErrorMessage(inputType, { inlineError, configError }) {
...
+  }
+
+  const isOptionalWithNull =
+    Array.isArray(field.jsonType) &&
+    // @TODO should also check the "oneOf" directly looking for "null" option
+    // but we don't have direct access at this point. Otherwise
+    // the JSON Schema validator will fail because setting jsonType is not enough.
+    field.jsonType.includes('null');
+  if (isOptionalWithNull) {
+    return [...optionsBroken, null];
```
sandrina-p, line 99: FYI I've created a dev release `0.4.1-dev.20230702181843` which includes this and #23. I'm testing it in our [internal repo](https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/17915)

src/yupSchema.js
brennj: Trying to parse my head around this right now 👍

src/yupSchema.js
sandrina-p: After all this `null` will always be necessary similar to `""`, otherwise our empty fields: Look at the [Radio demo](https://github.com/remoteoss/json-schema-form/assets/14869087/e431bea4-33bb-42fa-a3dd-589fe13fc9e2). (I spot this through our internal Storybook tests.

Changing the code (and code comment) now.

sandrina-p: Update: Done: [93fbd2e](https://github.com/remoteoss/json-schema-form/pull/18/commits/93fbd2eeef1239615909858f4e7a1700bc0c75ca). I've released a new version `0.4.1-dev.20230703082439` from #23.

brennj (commented):
Still going through this - code looks fine for now but also going through the failures we have internally against the pipeline

src/yupSchema.js
dilvane: I was wondering if we could take advantage of the `.transform()`, at least for the empty string case, to try to consider it a "Required field" error.

src/yupSchema.js
```diff
@@ -74,9 +78,65 @@ function getRequiredErrorMessage(inputType, { inlineError, configError }) {
...
+    Array.isArray(field.jsonType) &&
+    // @TODO should also check the "oneOf" directly looking for "null"
+    // option but we don't have direct access at this point.
+    // Otherwise the JSON Schema validator will fail as explained in PR#18
+    field.jsonType.includes('null');
+  if (isOptionalWithNull) {
+    return [...optionsBroken, null];
+  }
+
+  return optionsBroken;
```
brennj, line 108: based on my confusion on the internal pipelines failures. What are the instances where a null is not required?

brennj: fwiw i did test this by just always including `null` and `''` in the `optionsBroken` and it works .... but am I fully convinced its the way? no

src/yupSchema.js
```diff
@@ -74,9 +78,65 @@ function getRequiredErrorMessage(inputType, { inlineError, configError }) {
 
 const getJsonTypeInArray = (jsonType) =>
   Array.isArray(jsonType)
-    ? jsonType.find((val) => val !== 'null') // eg ["string", "null"] // optional fields - get the lead type.
+    ? jsonType.find((val) => val !== 'null') // eg ["string", "null"] // optional fields - get the head type.
     : jsonType; // eg "string"
 
+const getOptionsAllowed = (field) => {
+  const onlyValues = field.options?.map((option) => option.value);
```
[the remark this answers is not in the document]
sandrina-p: excellent question. The `disabled` is... yet another custom keyword, irrelevant for JSON Schemas (yes, it should be `x-jsf-presentation.disabled`). JSON Schemas do not have the concept of "disabled" fields/options. So for now, we'll leave it as it is - a normal option, even though technically speaking it shouldn't be allowed.

src/yupSchema.js
sandrina-p: > What are the instances where a null is not required?

Do you mean "allowed" instead of "required" right?

- In the ideal world: null is never allowed unless the field explicitly has an option with `const: null`.
- Nowadays: It must be always allowed, in the 3 scenarios:
  - required fields - allowed so the `.required()` works correctly saying "Required field" instead of "Option null is not valid".
  - optional fields (conventional)  - allowed so that we don't break existing integrations that send null when the field is empty.
  - optional fields (null) -  allowed (correct) because it has an option with `const: null`.

brennj: ok cool, so in that case .. for this one:

> optional fields (conventional) - allowed so that we don't break existing integrations that send null when the field is empty.

we either want to always include it or have a way to check when its conditionally required and don't include when it doesn't match that condition?

sandrina-p: Looking at our internal APIs, revealed a demo that is failing: [nested conditional fields](https://json-schema-form.vercel.app/?path=/story/combinations--nested-conditional-fields) - [local dev link](http://localhost:6007/?path=/story/combinations--nested-conditional-fields).

P.S. This is not a "nested conditional field". it's a "field with nested conditionals" - the field is not nested, it's normal. The conditional is the one that it's nesting. English is hard :sweat_smile: 

image

This error happens because the JSON Schema is wrong — `state` is an optional field. At Remote, as we treat empty fields as `null`, the field is missing an option with `const: null`. 

The same bugs happens in the other demos (and probably in our internal JSON Schemas too). If this json schema was a real one, our Remote API would fail with the typical "expected a string, got null" error.

For this reason, you are right, we'll need to keep allowing `null` always for now. And later on, create a migration-strategy:
- 1. in JSF log a warning when we spot these nulls as exceptions, as they should be an error instead.
- 2. in JSF add a feature-flag to allow devs to toggle it and start using the new version without these exceptions.
- 3. Use it ourselves to spot our incorrect json schemasand fix our own UI Form integration.

src/yupSchema.js
sandrina-p: Good point @dilvane , I'll look into `.transform()`. It might be easier to it instead, as we can use it in the future to also log warnings during the deprecation-phase. 💡

sandrina-p: @dilvane how exactly would you use `.transform()`? Whatever it returns, it must be valid for `.one()`. I would need a pseudo-code/PoC here from you pls. [screenshot](https://github.com/remoteoss/json-schema-form/assets/14869087/0b99cf19-0fb7-40e8-aa60-596dea6ce68c)

dilvane: @sandrina-p I was thinking something like this:

```js
.transform((value) => {
   return value === '' ? undefined : value;
})
```

So in the `getOptionsAllowed` function there is no need to add `` as allowed option;

src/helpers.js
```diff
@@ -107,9 +107,18 @@ function checkIfConditionMatches(node, formValues, formFields) {
...
     }
 
-    const { inputType } = getField(name, formFields);
-
-    return validateFieldSchema({ ...currentProperty, inputType, required: true }, value);
+    const field = getField(name, formFields);
+
+    return validateFieldSchema(
+      {
+        options: field.options,
```
sandrina-p, line 114: note: This solves a bug where conditionals with select fields weren't working. For example, the JSF would break on the following JSON Schema because it wouldn't be able to process the `if.properties.browser` correctly.

```json
{
  "allOf": [
    {
      "if": {
        "properties": {
          "browser": {
            "pattern": "."
          }
        },
        "required": [
          "browser"
        ]
      },
      "then": {},
      "else": {
        "properties": {
          "model": false
        }
      }
    }
  ],
  "properties": {
    "browser": {
      "oneOf": [
        {
          "const": "chr",
          "title": "Chrome"
        },
        {
          "const": "ff",
          "title": "Firefox"
        }
      ],
      "title": "Browser",
      "type": "string",
      "x-jsf-presentation": {
        "inputType": "select"
      }
    },
    "model": {
      "maxLength": 1000,
      "title": "Browser version",
      "type": "string",
      "x-jsf-presentation": {
        "inputType": "textarea"
      }
    }
  }
}
```

image

src/yupSchema.js
sandrina-p: Ahh, I see, let me try it 👀

sandrina-p: It worked nicely!! https://github.com/remoteoss/json-schema-form/pull/18/commits/6c7c1d7b4bec7fb2ec70aaae99f686f1db78da5c Shipping a new version to test in our FE codebase.

brennj:
Approving this one based on passing pieplines internally

brennj (approved):

src/yupSchema.js
```diff
@@ -34,8 +34,45 @@ const validateOnlyStrings = string()
...
+
+          [2] The null also needs to be always allowed for the same reason.
+              Some consumers (eg Remote) expect empty optional fields to be sent as "null"
+              even when their JSON Schemas are not created correctly (missing an option with const: null).
+              This will allow the JSF to still mark the field as valid (false positive)
+              and let the JSON Schema validator fail.
+
+          We'd need to implement a feature_flag/transition deprecation warning
+          to give devs the time to adapt their integrations before we fix this behavior.
+          Check the PR#18 and tests for more details.
```
sandrina-p, line 70: @gabrielseco @brennj As promised, I've recorded a [public loom](https://www.loom.com/share/3518f2ed732d4699b7ca83f3b220804b) (and added it to the PR description) explaining this workaround with examples.

Hopefully, it will help you better understand/explain this in the future when we come back to this to and create the migration phase to stop accepting `""` and `null` when we shouldn't.

PR #28, opened 2023-07-07
WIP: Advanced validation proof of concept

brennj:
WIP WIP WIP


src/tests/jsonLogicFixtures.js
```diff
@@ -0,0 +1,1098 @@
...
+          properties: {
+            field_c: false,
+          },
+        },
+      },
+    ],
+  },
+};
+
+export const schemaWithMultipleComputedValueChecks = {
```
sandrina-p, line 445: Interesting case. Here I'd expect the root field_c to have `'x-jsf-requiredValidations` declared even if in the conditional it it's disallowed (the `then ... false` bit). Which doesn't make sense as it's contradictory.

Perhaps a sanity test showing that impossible case would be useful for future reference.

src/tests/jsonLogicFixtures.js
```diff
@@ -0,0 +1,1098 @@
...
+        },
+      },
+    },
+    computedValues: {
+      a_times_two: {
+        rule: {
+          '*': [{ var: 'field_a' }, 2],
+        },
+      },
+      mod_by_five: {
```
sandrina-p, line 473: naming: I noticed multiple places where the naming doesn't make sense and I can foresee this happening in the future. Perhaps we'll need to document guidelines on how to name these rules. Let me give you some examples:

- Incomplete naming:
  - This one should be `b_mod_by_five`
  - At `schemaWithIfStatementWithComputedValuesAndValidationChecks`, the `greater_than_b` should be `a_greater_than_b` (or even `a_>_b`)
- Desbribe the action, not the outcome
- At `schemaWithGreaterThanChecksForThreeFields` the `require_c` should be `a_greater_than_b`. Then based on this validation, we apply wtv outcome it's needed (in that example, the c gets required)

There are other cases through the examples, but I think you go the point. Does this make sense to you? :)

src/tests/jsonLogic.test.js
```diff
@@ -0,0 +1,649 @@
...
+    });
+
+    it('Handle having a true condition with both validations and computedValue checks', () => {
+      const { handleValidation } = createHeadlessForm(
+        schemaWithIfStatementWithComputedValuesAndValidationChecks,
+        { strictInputType: false }
+      );
+      expect(handleValidation({ field_a: 1, field_b: 1 }).formErrors).toEqual(undefined);
+      expect(handleValidation({ field_a: 10, field_b: 20 }).formErrors).toEqual(undefined);
+      expect(handleValidation({ field_a: 10, field_b: 9 }).formErrors).toEqual({
```
sandrina-p, line 418: todo: In this type of tests (with conditional fields) we should also test the visibility for field_c is `isVisible: false`.

tip: You may want to use use the util I created on the other test file called `getField()`

src/tests/jsonLogic.test.js
```diff
@@ -0,0 +1,649 @@
...
+
+    it('computedAttribute test that minimum, errorMessages.minimum is working', () => {
+      const { handleValidation } = createHeadlessForm(
+        aConditionallyAppliedComputedAttributeMinimum,
+        {
+          strictInputType: false,
+        }
+      );
+      expect(handleValidation({ field_a: 20, field_b: 1 }).formErrors).toEqual({
+        field_b: 'use 10 or more',
```
sandrina-p, line 482: incredible

src/tests/jsonLogic.test.js
```diff
@@ -0,0 +1,649 @@
...
+    it('Apply a conditional computed Attrbute value', () => {
+      const { fields, handleValidation } = createHeadlessForm(
+        aConditionallyAppliedComputedAttributeValue,
+        {
+          strictInputType: false,
+        }
+      );
+
+      //TODO: this should fail because we have a const: 10, it should have an error from yup saying only 10 is accepted.
+      expect(handleValidation({ field_a: 20, field_b: 1 }).formErrors).toEqual();
```
sandrina-p, line 497: just leaving a comment to highlight this bug that you spot

src/tests/jsonLogicFixtures.js
```diff
@@ -0,0 +1,1098 @@
...
+            rule: {
+              '>': [{ var: 'child' }, 10],
+            },
+          },
+        },
+      },
+    },
+    field_b: {
+      type: 'number',
+      'x-jsf-requiredValidations': ['validation_parent', 'peek_to_nested'],
```
sandrina-p, line 745: question: Why is the `peek_to_nested` here if it's about `field_a.child` which has no influence under `field_b`?

It makes me wonder if a validation at `requiredValidations` MUST always have something in common with the field itself.

src/tests/jsonLogicFixtures.js
```diff
@@ -0,0 +1,1098 @@
...
+  properties: {
+    field_a: {
+      type: 'object',
+      'x-jsf-presentation': {
+        inputType: 'fieldset',
+      },
+      properties: {
+        child: {
+          type: 'number',
+          'x-jsf-requiredValidations': ['child_greater_than_10'],
```
sandrina-p, line 728: question: I tried the following variant where I use `peek_to_nested` here:

<details>

```json
export const twoLevelsOfJSONLogicSchema = {
  properties: {
    field_a: {
      type: 'object',
      'x-jsf-presentation': {
        inputType: 'fieldset',
      },
      properties: {
        child: {
          type: 'number',
          'x-jsf-requiredValidations': ['child_greater_than_10', 'peek_to_nested'],
        },
      },
      required: ['child'],
      'x-jsf-logic': {
        validations: {
          child_greater_than_10: {
            errorMessage: 'Must be smaller than 50!',
            rule: {
              '<': [{ var: 'child' }, 50],
            },
          },
        },
      },
    },
    field_b: {
      type: 'number',
      'x-jsf-requiredValidations': ['validation_parent'],
    },
  },
  'x-jsf-logic': {
    validations: {
      validation_parent: {
        errorMessage: 'Must be greater than 10!',
        rule: {
          '>': [{ var: 'field_b' }, 10],
        },
      },
      peek_to_nested: {
        errorMessage: 'child must be greater than 15!',
        rule: {
          '>': [{ var: 'field_a.child' }, 15],
        },
      },
    },
  },
  required: ['field_a', 'field_b'],
};
```

</details>

And then run the respective test `"Validate a field and a nested field together"`. As expected the test failed:

```
    JSON Schema invalid! Error: Validation "peek_to_nested" required for "child" doesn't exist.
```

I think we could have an extra test to ensure this will always fail for sanity check. We already have for root fields, but not for fieldsets.

src/tests/jsonLogic.test.js
```diff
@@ -0,0 +1,649 @@
...
+        handleValidation({ field_array: [{ array_item: 2 }, { array_item: 1 }] }).formErrors
+      ).toEqual({
+        field_array: [undefined, { array_item: 'Must be divisible by two' }],
+      });
+      expect(
+        handleValidation({ field_array: [{ array_item: 2 }, { array_item: 2 }] }).formErrors
+      ).toEqual(undefined);
+    });
+
+    it('Validating a single item in an array should work', () => {
```
sandrina-p, line 635: supreme 🔥

sandrina-p (commented):
I reviewed the tests and respective fixtures. Will review the actual code later, probably only on Monday.

PR #31, opened 2023-07-24
fix(date) - fix breaking field when not passing the property format

gabrielseco:
1. Description
We have detected internally that some of our forms stopped working as we were not passing the property format to some data fields.

This MR fixes this and adds some tests to prevent them from failure in the feature

2. How to test this

Instructions to how to test this internally are [here](https://remote-com.slack.com/archives/C03RWJF31U2/p1690190265877309?thread_ts=1689957612.695479&cid=C03RWJF31U2)

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -1117,6 +1117,29 @@ describe('createHeadlessForm', () => {
       });
     });
 
+    it('support format date with minDate and maxDate without the format property', () => {
+      const schemaFormatDate = {
+        properties: {
+          birthdate: {
+            title: 'Birthdate',
+            type: 'string',
+            'x-jsf-presentation': {
+              inputType: 'date',
+              maxDate: '2022-03-01',
+              minDate: '1922-03-01',
+            },
+          },
+        },
+      };
+
+      const { handleValidation } = createHeadlessForm(schemaFormatDate);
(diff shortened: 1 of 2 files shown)


src/tests/createHeadlessForm.test.js
```diff
@@ -1117,6 +1117,29 @@ describe('createHeadlessForm', () => {
...
     });
 
+    it('support format date with minDate and maxDate without the format property', () => {
+      const schemaFormatDate = {
+        properties: {
+          birthdate: {
+            title: 'Birthdate',
+            type: 'string',
+            'x-jsf-presentation': {
+              inputType: 'date',
```
sandrina-p, line 1127: @gabrielseco, if I understood correctly, this solution is a workaround to an invalid JSON Schema that does not have `format: date`, right? 
If so, that means that our JSON Schemas are not correct and the data received is not being validated correctly in the BE (because the format is missing, so any string is valid).

In other words, this PR isn't the right solution. Instead, we should fix the incorrect JSON Schemas we have internally.

Does this make sense to you?

gabrielseco: Yes, I thought that maybe the format property was optional before my change and was not present in all json-schemas.
> this solution is a workaround to an invalid JSON Schema that does not have format: date, right?

Let's close this MR, then

PR #33, opened 2023-08-08
Allow deeply nested if property checks

brennj:
Internal conversation https://remote-com.slack.com/archives/C02HTN0LY02/p1691493090023339

When we have a if check that needs to check two levels deep, it would break the `json-schema-form` library, this adds a check to `checkIfConditionMatches` to make it more capable to do understand these property types.

Theres enough logic to justify its own file, and I started some dedicated tests.


rylek90 (approved):
Looks fine, thanks for adding tests 👀

PR #34, opened 2023-08-16
[JSON-Logic] Part 0: Allow consts to support single values only

brennj:
Part 1 of merging JSON Logic schemas in, but this is really its own standalone piece of work.

Today if we have a `const` value in a schema, it doesn't take into account validations to only allow that value.

With this PR, it adds that functionality to cause an error, also where you can define the custom error using `x-jsf-errorMessage.const`.

Please see the tests for sample schemas on how to make this work.

### DIFF
--- a/src/tests/const.test.js
+++ b/src/tests/const.test.js
@@ -1,78 +1,86 @@
 import { createHeadlessForm } from '../createHeadlessForm';
 
-it('Should work for number', () => {
-  const { handleValidation } = createHeadlessForm(
-    {
-      properties: {
-        ten_only: { type: 'number', const: 10 },
-      },
-    },
-    { strictInputType: false }
-  );
-  expect(handleValidation({}).formErrors).toEqual(undefined);
-  expect(handleValidation({ ten_only: 1 }).formErrors).toEqual({
-    ten_only: 'The only accepted value is 10.',
-  });
-  expect(handleValidation({ ten_only: 10 }).formErrors).toEqual(undefined);
-});
-
-it('Should work for text', () => {
-  const { handleValidation } = createHeadlessForm(
-    {
-      properties: {
-        hello_only: { type: 'string', const: 'hello' },
-      },
-    },
-    { strictInputType: false }
-  );
-  expect(handleValidation({}).formErrors).toEqual(undefined);
-  expect(handleValidation({ hello_only: 'what' }).formErrors).toEqual({
-    hello_only: 'The only accepted value is hello.',
-  });
-  expect(handleValidation({ hello_only: 'hello' }).formErrors).toEqual(undefined);
-});
-
-it('Should work for a conditionally applied const', () => {
-  const { handleValidation } = createHeadlessForm(
-    {
-      properties: {
-        answer: { type: 'string', oneOf: [{ const: 'yes' }, { const: 'no' }] },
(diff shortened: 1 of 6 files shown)


johnstonbl01 (dismissed):

src/yupSchema.js
```diff
@@ -304,6 +304,19 @@ export function buildYupSchema(field, config) {
     );
   }
 
+  function withConst(yupSchema) {
+    return yupSchema.test(
+      'isConst',
+      errorMessage.const ??
```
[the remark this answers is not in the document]
sandrina-p: We have just [one sentence](https://json-schema-form.vercel.app/?path=/docs/api-reference-json-schema--custom-keywords#x-jsf-errormessage) about that it the docs. Probably a table or diagram could explain it better.

src/tests/const.test.js
```diff
@@ -0,0 +1,78 @@
+import { createHeadlessForm } from '../createHeadlessForm';
+
+it('Should work for number', () => {
```
sandrina-p, line 3: question: Is this the moment we start to break our tests into smaller files? If so this should be a `describe('validations: const', ())` with all the `it` inside it.

src/tests/const.test.js
```diff
@@ -0,0 +1,78 @@
...
+        ten_only: { type: 'number', const: 10 },
+      },
+    },
+    { strictInputType: false }
+  );
+  expect(handleValidation({}).formErrors).toEqual(undefined);
+  expect(handleValidation({ ten_only: 1 }).formErrors).toEqual({
+    ten_only: 'The only accepted value is 10.',
+  });
+  expect(handleValidation({ ten_only: 10 }).formErrors).toEqual(undefined);
```
sandrina-p, line 16: todo: It's missing to test the `ten_only: null` to cover the "bug" `@BUG RMT-518` (check other comments in the `createHeadlessForm.test.js`)

```js
  // null is also considered valid until we fix @BUG RMT-518
  // Expectation: To fail with error "The only accepted value is 10."
```

src/tests/const.test.js
brennj: yes it is, no more massive test files 😅 , will fix this.

src/tests/const.test.js
brennj: Added!

brennj:
@sandrina-p are you the admin on this repo? Can we change it so that new commits dismiss old reviews?

if you or @johnstonbl01 can give this a "reapproval" i think we can merge now, green internally now!

src/tests/const.test.js
```diff
@@ -0,0 +1,86 @@
+import { createHeadlessForm } from '../createHeadlessForm';
```
sandrina-p, line 1: > @sandrina-p are you the admin on this repo? Can we change it so that new commits dismiss old reviews?

> if you or @johnstonbl01 can give this a "reapproval" i think we can merge now, green internally now!


I hate github for not allowing threads in "standalone" comments, so I'm creating this new comment:

sandrina-p: Yes, I was the only admin, I changed that now, pls double-check. I also added a branch rule protection to `main` to require re-approvals after commits.

brennj: Looks good, thank you!

PR #36, opened 2023-08-22
[JSON Logic] Part 2: Computed Attribute skeleton

brennj:
Requires: https://github.com/remoteoss/json-schema-form/pull/34 and https://github.com/remoteoss/json-schema-form/pull/35 for full context and to work correctly. Must only be merged when they are.

This is a breakdown of https://github.com/remoteoss/json-schema-form/pull/28 to make it more digestible for review.

**Add in x-jsf-logic-computedAttrs (with reference to values defined in x-jsf-logic.computedValues) support.**

- Ensures when there is a `x-jsf-logicComputedAttrs` are present, it reads from `x-jsf-logic.computedValues` for that computed value.
- From this PR, any plain string based single value is supported. (Where you need titles, etc, inline rules, they will come in following PRs)
- Forced values are supported for this.
- Adds a test to show how this works for basic values (const, default)

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -574,8 +574,8 @@ export function yupToFormErrors(yupError) {
  * @param {JsfConfig} config - jsf config
  * @returns {Function(values: Object): { YupError: YupObject, formErrors: Object }} Callback that returns Yup errors <YupObject>
  */
-export const handleValuesChange = (fields, jsonSchema, config) => (values) => {
-  updateFieldsProperties(fields, values, jsonSchema);
+export const handleValuesChange = (fields, jsonSchema, config, logic) => (values) => {
+  updateFieldsProperties(fields, values, jsonSchema, logic);
 
   const lazySchema = lazy(() => buildCompleteYupSchema(fields, config));
   let errors;
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -480,6 +480,7 @@ export function extractParametersFromNode(schemaNode) {
   return omitBy(
     {
       const: node.const,
+      ...(node.const && node.default ? { value: node.const } : {}),
       label: node.title,
       readOnly: node.readOnly,
       ...(node.deprecated && {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -280,7 +280,7 @@ export function processNode({
   });
 
   if (node.if) {
-    const matchesCondition = checkIfConditionMatches(node, formValues, formFields);
+    const matchesCondition = checkIfConditionMatches(node, formValues, formFields, logic);
     // BUG HERE (unreleated) - what if it matches but doesn't has a then,
     // it should do nothing, but instead it jumps to node.else when it shouldn't.
     if (matchesCondition && node.then) {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -179,7 +179,7 @@ export function getPrefillValues(fields, initialValues = {}) {
  * @param {Object} node - JSON-schema node
  * @returns
  */
-function updateField(field, requiredFields, node, formValues) {
+function updateField(field, requiredFields, node, formValues, validations, config) {
   // If there was an error building the field, it might not exist in the form even though
   // it can be mentioned in the schema so we return early in that case
   if (!field) {
@@ -226,6 +226,18 @@ function updateField(field, requiredFields, node, formValues) {
       }
     });
 
+  if (field.getComputedAttributes) {
+    const computedFieldValues = field.getComputedAttributes({
+      field,
+      isRequired: fieldIsRequired,
+      node,
+      formValues,
+      config,
+      validations,
+    });
+    updateValues(computedFieldValues);
+  }
+
   // If field has a calculateConditionalProperties closure, run it and update the field properties
   if (field.calculateConditionalProperties) {
     const newFieldValues = field.calculateConditionalProperties(fieldIsRequired, node);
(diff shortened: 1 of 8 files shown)


src/helpers.js
```diff
@@ -442,6 +456,15 @@ export function extractParametersFromNode(schemaNode) {
   const presentation = pickXKey(schemaNode, 'presentation') ?? {};
   const errorMessage = pickXKey(schemaNode, 'errorMessage') ?? {};
   const requiredValidations = schemaNode['x-jsf-logic-validations'];
+  const computedAttributes = schemaNode['x-jsf-logic-computedAttrs'];
+
+  // This is when a forced value is computed.
+  const decoratedComputedAttributes = {
```
johnstonbl01, line 462: **Nitpick:** Could this be done in its own function? There's a lot of conditional logic happening in these 4 lines?

src/jsonLogic.js
```diff
@@ -106,3 +106,33 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
+
+function handleComputedAttribute(validations, formValues, parentID, name) {
+  return ([key, value]) => {
+    if (key === 'const')
+      return [
+        key,
+        validations.getScope(parentID).evaluateComputedValueRuleForField(value, formValues, name),
+      ];
+
+    if (typeof value === 'string') {
```
johnstonbl01, line 131: Just want to confirm - subsequent PRs (honestly have been waiting so long to use PR instead of MR again 😂 ) will include additional logic for this function, so that we do not need to handle other cases where `key !== 'const'` or `typeof value === 'string'` - is that correct? If it is, no worries. If not, then do we need to guard against values we don't recognize, or will it silently fail?

johnstonbl01 (dismissed):

src/jsonLogic.js
johnstonbl01: Confirmed when I looked at Part 3. :)

src/tests/jsonLogic.test.js
```diff
@@ -210,4 +211,16 @@ describe('cross-value validations', () => {
...
+    it('field_b is field_a * 2', () => {
+      const { fields } = createHeadlessForm(schemaWithComputedAttributes, {
+        strictInputType: false,
+        initialValues: { field_a: 2 },
+      });
+      const fieldB = fields.find((i) => i.name === 'field_b');
+      expect(fieldB.default).toEqual(4);
+      expect(fieldB.value).toEqual(4);
+    });
+  });
```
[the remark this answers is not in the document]
dilvane: Please ignore me if this is not part of the PR. Thanks!

src/jsonLogic.js
```diff
@@ -106,3 +106,33 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
+    return attributes;
+  };
+}
+
+function handleComputedAttribute(validations, formValues, parentID, name) {
+  return ([key, value]) => {
+    if (key === 'const')
+      return [
+        key,
+        validations.getScope(parentID).evaluateComputedValueRuleForField(value, formValues, name),
```
dilvane, line 128: Looking at the signature of the `evaluateComputedValueRuleForField` it appears that the last parameter, `name`, is not expected.

src/tests/jsonLogic.test.js
```diff
@@ -210,4 +211,16 @@ describe('cross-value validations', () => {
...
+
+  describe('Derive values', () => {
+    it('field_b is field_a * 2', () => {
+      const { fields } = createHeadlessForm(schemaWithComputedAttributes, {
+        strictInputType: false,
+        initialValues: { field_a: 2 },
+      });
+      const fieldB = fields.find((i) => i.name === 'field_b');
+      expect(fieldB.default).toEqual(4);
+      expect(fieldB.value).toEqual(4);
```
sandrina-p, line 223: todo/bug: I didn't remember this before, but I'm concerned with this new `value` key because it might conflict with our internal codebase (`composeField*` fns) where value is already used for other legacy things.
Hopefully, I'm wrong, but could you please open a PR with a demo to test that it integrates well out of the box? (related [linear issue](https://linear.app/remote/issue/RMT-18))

src/jsonLogic.js
brennj: this will come soon (error handling) but i'll remove for now

src/tests/jsonLogic.test.js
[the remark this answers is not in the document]
brennj: good spot, i thought we didn't need to pass down `logic` until a further PR, but lets just do it here to prevent this error.

src/helpers.js
brennj: Updated

src/tests/jsonLogic.test.js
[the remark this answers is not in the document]
brennj: Updated

src/jsonLogic.js
brennj: Updated

src/tests/jsonLogic.test.js
brennj: On it.

brennj: Internal MR opened

sandrina-p: will review it later today!

sandrina-p (approved):

PR #37, opened 2023-08-23
[JSON Logic] Part 3: Computed string based values

brennj:
Must be merged after https://github.com/remoteoss/json-schema-form/pull/36

**Add in x-jsf-logic-computedAttrs (with reference to values defined in x-jsf-logic.computedValues) support.**

- Supports dynamic:
  - `x-jsf-errorMessage.minimum/maximum`
  - `title`
  - `description`
  -  `x-jsf.presentation.statement.description`

### DIFF
--- a/src/jsonLogic.js
+++ b/src/jsonLogic.js
@@ -146,6 +146,12 @@ export function yupSchemaWithCustomJSONLogic({ field, logic, config, id }) {
   const { parentID = 'root' } = config;
   const validation = logic.getScope(parentID).validationMap.get(id);
 
+  if (validation === undefined) {
+    throw Error(
+      `[json-schema-form] json-logic error: "${field.name}" required validation "${id}" doesn't exist.`
+    );
+  }
+
   return (yupSchema) =>
     yupSchema.test(
       `${field.name}-validation-${id}`,
@@ -182,6 +188,21 @@ function replaceHandlebarsTemplates({
 }) {
   if (typeof toReplace === 'string') {
     return toReplace.replace(HANDLEBARS_REGEX, (match, key) => {
+      return logic.getScope(parentID).applyComputedValueInField(key.trim(), formValues, fieldName);
+    });
+  } else if (typeof toReplace === 'object') {
+    const { value, ...rules } = toReplace;
+
+    if (Object.keys(rules).length > 1 && !value) {
+      throw Error('Cannot define multiple rules without a template string with key `value`.');
+    }
+
+    const computedTemplateValue = Object.entries(rules).reduce((prev, [key, rule]) => {
+      const computedValue = logic.getScope(parentID).evaluateValidation(rule, formValues);
+      return prev.replaceAll(`{{${key}}}`, computedValue);
+    }, value);
+
+    return computedTemplateValue.replace(/\{\{([^{}]+)\}\}/g, (match, key) => {
       return logic.getScope(parentID).applyComputedValueInField(key.trim(), formValues, fieldName);
     });
   }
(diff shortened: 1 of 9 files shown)


src/jsonLogic.js
```diff
@@ -107,32 +109,106 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
+      const computedValue = validations.getScope(parentID).evaluateValidation(rule, formValues);
+      return prev.replaceAll(`{{${key}}}`, computedValue);
+    }, value);
+
+    return computedTemplateValue.replace(/\{\{([^{}]+)\}\}/g, (match, key) => {
+      return validations
+        .getScope(parentID)
+        .evaluateComputedValueRuleForField(key.trim(), formValues, fieldName);
+    });
+  }
```
johnstonbl01, line 138: Does this need an `else` that returns `value` or something?

src/jsonLogic.js
```diff
@@ -107,32 +109,106 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
+        config,
+        validations
+      ),
+    };
   };
 }
 
 function handleComputedAttribute(validations, formValues, parentID, name) {
   return ([key, value]) => {
+    if (key === 'description')
```
johnstonbl01, line 163: Can we add `{}` to all the conditional blocks here for consistency, please? 🙏

johnstonbl01 (changes requested):

src/jsonLogic.js
```diff
@@ -107,32 +109,106 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
+  parentID,
+  name: fieldName,
+}) {
+  if (typeof toReplace === 'string') {
+    return toReplace.replace(/\{\{([^{}]+)\}\}/g, (match, key) => {
+      return validations
+        .getScope(parentID)
+        .evaluateComputedValueRuleForField(key.trim(), formValues, fieldName);
+    });
+  } else if (typeof toReplace === 'object') {
```
dilvane, line 125: Do you have a case where the `toReplace` is an object? Looking at the `schemaWithComputedAttributesAndErrorMessages` even `x-jsf-errorMessage` doesn't enter here.

src/jsonLogic.js
```diff
@@ -107,32 +109,106 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
+  validations,
+  formValues,
+  parentID,
+  name: fieldName,
+}) {
+  if (typeof toReplace === 'string') {
+    return toReplace.replace(/\{\{([^{}]+)\}\}/g, (match, key) => {
+      return validations
+        .getScope(parentID)
+        .evaluateComputedValueRuleForField(key.trim(), formValues, fieldName);
```
sandrina-p, line 123: nit: Following my comments from #34, I'm now more confident that this fn should be called `apply*` or `get*` because we are really returning that value, not just "evaluating" true/false

src/jsonLogic.js
sandrina-p: Same with tests, none of the tests cover this `else` branch

src/jsonLogic.js
sandrina-p: I'm surprised eslint isn't setup to enforce [multi-line](https://eslint.org/docs/latest/rules/curly#multi-line) 🤔 .

Could you add to `.eslintrc` the following rule?

```
    "curly": ["error", "multi-line"],
```

src/jsonLogic.js
```diff
@@ -107,32 +109,106 @@ export function yupSchemaWithCustomJSONLogic({ field, validations, config, id })
...
 
+function replaceHandlebarsTemplates({
+  value: toReplace,
+  validations,
+  formValues,
+  parentID,
+  name: fieldName,
+}) {
+  if (typeof toReplace === 'string') {
+    return toReplace.replace(/\{\{([^{}]+)\}\}/g, (match, key) => {
```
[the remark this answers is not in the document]
brennj: Done!

src/jsonLogic.js
brennj: added the value to return in the last line, good call

src/jsonLogic.js
brennj: I was very confused for _why_ this wasn't being called but it was a mistake on my part. This actually goes with the inline rules code. I'll shift it to the appropriate PR. It will be removed for this PR.

src/jsonLogic.js
brennj: this is done

src/jsonLogic.js
johnstonbl01: Appears that it's been changed to `applyComputedValueInField`, so resolving this one.

johnstonbl01 (approved):

sandrina-p (dismissed):
Didn't find anything else odd, let's go p4!

dilvane (approved):

brennj:
lets goo

PR #38, opened 2023-08-24
[JSON Logic] Part 4: Computed Attribute error handling

brennj:
Continuing the train:

- Adds error checking for inline rules (which isn't in place yet here, will be in next MR)
- Adds error checking for missing validation ids and computedAttr ids 
- Adds tests and schemas to verify these.


src/tests/jsonLogic.test.js
```diff
@@ -82,6 +89,82 @@ describe('cross-value validations', () => {
...
+
+    it('Should throw when theres an inline computed ruleset with no value.', () => {
+      createHeadlessForm(schemaWithMissingValueInlineRule, { strictInputType: false });
+      expect(console.error).toHaveBeenCalledWith(
+        'JSON Schema invalid!',
+        Error('Cannot define multiple rules without a template string with key `value`.')
+      );
+    });
+
+    it('On x-jsf-logic-computedAttrs, error if theres a value that does not exist.', () => {
```
sandrina-p, line 125: nit: Did you write these tests on a different day? Their description are not consistent compared to the ones above 😅  More similar: "Should throw on x-jsf-logic-computedAttrs, if there's a value that does not exist."

Or may personal favorite which cuts "should", and uses "given" instead of "if": "On x-jsf-logic-computedAttrs, throw error given a value that doesn't exist"

src/tests/jsonLogicFixtures.js
```diff
@@ -178,6 +231,54 @@ export const schemaWithInlineRuleForComputedAttributeWithoutCopy = {
...
+    field_a: {
+      type: 'number',
+      'x-jsf-logic-computedAttrs': {
+        description: `this doesn't exist {{iDontExist}}`,
+      },
+    },
+  },
+};
+
 export const schemaWithInlineRuleForComputedAttributeWithOnlyTheRule = {
```
sandrina-p, line 282: todo: this var is not used anywhere 🤔

src/jsonLogic.js
```diff
@@ -125,6 +144,9 @@ function replaceHandlebarsTemplates({
   } else if (typeof toReplace === 'object') {
     const { value, ...rules } = toReplace;
 
+    if (Object.keys(rules).length > 1 && !value)
+      throw Error('Cannot define multiple rules without a template string with key `value`.');
+
```
sandrina-p, line 149: eslint: missing the brace `if { ... }` as we talked in the other PR

src/jsonLogic.js
```diff
@@ -212,3 +234,71 @@ function handleNestedObjectForComputedValues(values, formValues, parentID, valid
...
+              sampleEmptyObject,
+              (item) =>
+                `"${item.var}" in inline rule in property "${fieldName}.x-jsf-logic-computedAttrs.${key}" does not exist as a JSON schema property.`
+            );
+          });
+        });
+    });
+}
+
+function checkRuleIntegrity(
```
sandrina-p, line 279: todo: A jsDoc with explanation + simple example across all these new functions would be very useful

src/jsonLogic.js
```diff
@@ -212,3 +234,71 @@ function handleNestedObjectForComputedValues(values, formValues, parentID, valid
...
+  rule,
+  id,
+  data,
+  errorMessage = (item) => `"${item.var}" in rule "${id}" does not exist as a JSON schema property.`
```
sandrina-p, line 283: tests: This scenario (default error) is not covered in the tests, right?

src/jsonLogic.js
```diff
@@ -212,3 +234,71 @@ function handleNestedObjectForComputedValues(values, formValues, parentID, valid
...
+) {
+  Object.values(rule ?? {}).map((subRule) => {
```
sandrina-p, line 285: todo: A code comment with a code snippet as example would be useful. If I understood correctly, for now, this fn only checks if the defined `var` points to existing fields.

src/jsonLogic.js
```diff
@@ -125,6 +144,9 @@ function replaceHandlebarsTemplates({
   } else if (typeof toReplace === 'object') {
     const { value, ...rules } = toReplace;
 
+    if (Object.keys(rules).length > 1 && !value)
```
sandrina-p, line 147: bug: What if the `value` does not exist? Then the code will break below at `prev.replaceAll`. For example:

<details>
```js
{
  properties: {
    field_a: {
      type: 'number',
      'x-jsf-logic-computedAttrs': {
        title: {
          // value: 'Hi {{rule}}',
          rule: {
            '+': [{ var: 'field_a' }, 4],
          },
        },
      },
    },
  },
}
```
</details>

sandrina-p: Ah, I saw now the Part 5 https://github.com/remoteoss/json-schema-form/pull/39/files includes more validations. Nevermind!

src/jsonLogic.js
```diff
@@ -212,3 +234,71 @@ function handleNestedObjectForComputedValues(values, formValues, parentID, valid
...
+      sample[key] = buildSampleEmptyObject(itemSchema);
+    } else {
+      sample[key] = true;
+    }
+  }
+
+  return sample;
+}
+
+function validateInlineRules(jsonSchema, sampleEmptyObject) {
```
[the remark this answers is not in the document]
sandrina-p: Update: Another thing I remember while reviewing Part 5. This code assumes it will run/validate the json schema after it's created/written. But at Remote we'll build them in a UI (form builder). I wonder if these validations will run too soon. 

For example, I'm still creating the logic pointers (i.e. `x-jsf-logic-validations`) and it's already complaining that the validation declaration (`x-jsf-logic.validations`) doesn't exist. Or the other way around. I build the declaration but it says the field and or point doesn't exist yet. I think that's a very valid scenario to silent the checkers/validations while I'm in "edit mode". But not sure. We'd need to actually test it to see how it behaves.

src/jsonLogic.js
```diff
@@ -44,12 +47,25 @@ function createValidationsScope(schema) {
 
   const validations = Object.entries(logic.validations ?? {});
   const computedValues = Object.entries(logic.computedValues ?? {});
+  const sampleEmptyObject = buildSampleEmptyObject(schema);
 
   validations.forEach(([id, validation]) => {
+    if (!validation.rule) {
+      throw Error(`Missing rule for validation with id of: "${id}".`);
```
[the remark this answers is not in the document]
brennj: Made some moves here, scoping everything with `[json-schema-form] json-logic error: something something`, does it make more sense? Open to more feedback on this one

src/tests/jsonLogicFixtures.js
brennj: this was my mistake was and should only be used later

src/jsonLogic.js
```diff
@@ -212,3 +234,71 @@ function handleNestedObjectForComputedValues(values, formValues, parentID, valid
...
+        }
+      } else {
+        checkRuleIntegrity(item, id, data);
+      }
+    });
+  });
+}
+
+function removeIndicesFromPath(path) {
+  const intermediatePath = path.replace(/\.\d+\./g, '.');
```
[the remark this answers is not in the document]
brennj: is the jsdoc ok here?

src/jsonLogic.js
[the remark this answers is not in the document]
brennj: Yeah, I'm all up for a configuration and this could be a good idea - but lets see it when we feel the pain of using it internally.

src/jsonLogic.js
brennj: Could you let me know if the jsdoc covers this point for you?

src/tests/jsonLogic.test.js
```diff
@@ -82,6 +89,82 @@ describe('cross-value validations', () => {
     });
   });
 
+  describe('Incorrectly written schemas', () => {
+    beforeEach(() => {
```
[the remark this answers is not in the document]
brennj: Nice, I added this to the top and extracted some shared utils in https://github.com/remoteoss/json-schema-form/pull/38/commits/df9b1a62416a06842240935ee4c86c882ec0e1e8.

src/tests/jsonLogic.test.js
```diff
@@ -82,6 +89,82 @@ describe('cross-value validations', () => {
...
+      jest.spyOn(console, 'error').mockImplementation(() => {});
+    });
+
+    afterEach(() => {
+      console.error.mockRestore();
+    });
+
+    it('Should throw when theres a missing rule', () => {
+      createHeadlessForm(schemaWithMissingRule, { strictInputType: false });
+      expect(console.error).toHaveBeenCalledWith(
```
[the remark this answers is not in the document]
brennj: this is a good call, i included this. I also just cleaned up the error code to use `test.each(cases)` because its all the same

src/jsonLogic.js
brennj: You're correct, i added two tests in https://github.com/remoteoss/json-schema-form/pull/38/commits/827ba1b8cd16f52f0c5e418fef9f322f94d7fa27 to account for this for more rule integrity checks.

src/tests/jsonLogic.test.js
brennj: i tried to clean this up with https://github.com/remoteoss/json-schema-form/pull/38/commits/81402ed910a066a3d3ba2a7c1dd58f808548deed. Another day of https://www.youtube.com/watch?v=JXFGy10b7Js, LMK if this could be better

src/jsonLogic.js
brennj: since this is inline rules, its gone from here but will be reintroduced later down in that relevant PR

src/tests/jsonLogic.test.js
```diff
@@ -82,6 +89,82 @@ describe('cross-value validations', () => {
...
+      createHeadlessForm(schemaWithComputedAttributeThatDoesntExist, {
+        strictInputType: false,
+      });
+      expect(console.error).toHaveBeenCalledWith(
+        'JSON Schema invalid!',
+        Error(`"iDontExist" computedValue in field "field_a" doesn't exist.`)
+      );
+    });
+
+    it('On x-jsf-logic-computedAttrs, error if theres a value that does not exist on a title.', () => {
```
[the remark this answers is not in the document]
brennj: it needs to be different because we use `label` for the title and `description` for description, i wanna be sure this still works, a separate test makes sense here imo

src/tests/jsonLogic.test.js
```diff
@@ -82,6 +89,82 @@ describe('cross-value validations', () => {
...
+    it('Should throw when theres a missing computed value', () => {
+      createHeadlessForm(schemaWithMissingComputedValue, { strictInputType: false });
+      expect(console.error).toHaveBeenCalledWith(
+        'JSON Schema invalid!',
+        Error('Missing rule for computedValue with id of: "a_plus_ten".')
+      );
+    });
+
+    it('Should throw when theres an inline computed ruleset with no value.', () => {
+      createHeadlessForm(schemaWithMissingValueInlineRule, { strictInputType: false });
```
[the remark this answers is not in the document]
brennj: yup this is gone for now but will come back :)

src/tests/jsonLogic.test.js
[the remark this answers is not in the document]
brennj: also, forgot to mention, this makes a good point that it isn't clear enough, i've switched (heh) to a switch statement to try and make the flow a bit more logical, because this was indeed a product of me piecing this function together as it made more sense. (see https://github.com/remoteoss/json-schema-form/pull/38/commits/02a0ff5aadf9ad15833f81427637cf61ef70033c)

src/jsonLogic.js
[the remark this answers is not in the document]
johnstonbl01: Agree with the sentiment of blocking the user in favor of more consistent data. I also think the internal UX of not running the validation while creating the schema is better, but that may be moot if the performance isn't an issue with re-running it constantly. Additionally, if performance or running the validation too frequently is an issue, we can optimize by debouncing, queuing, or possibly even off-loading some of the work to a service worker.

src/jsonLogic.js
```diff
@@ -203,3 +245,119 @@ function handleNestedObjectForComputedValues(values, formValues, parentID, logic
...
+ * referenced in the rules exist in the schema.
+ *
+ * Throws if any variable in a computed attribute rule does not exist.
+ *
+ * @param {Object} jsonSchema - The JSON schema object
+ * @param {Object} sampleEmptyObject - Sample empty object based on the schema
+ */
+function validateInlineRules(jsonSchema, sampleEmptyObject) {
+  const properties = (jsonSchema?.properties || jsonSchema?.items?.properties) ?? {};
+  Object.entries(properties)
```
johnstonbl01, line 293: Nit: I think this may be easier to read if it were done recursively?

src/jsonLogic.js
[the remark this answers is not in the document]
johnstonbl01: Nit: I know it's a simple Regex, but I am in favor of moving this to a reusable constant as well.

src/tests/jsonLogic.test.js
sandrina-p: there you go, green card eheh

src/jsonLogic.js
sandrina-p: yes totally 💯

src/jsonLogic.js
brennj: lots of small tweaks in f3b0895ec90890913004d538d9c2d3fb4e528ba0 for this, LMK!

src/jsonLogic.js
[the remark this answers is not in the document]
brennj: added it in its own var :)

src/jsonLogic.js
brennj: this is what I tried but it wasn't working as I was expecting:

```

function validateInlineRules(jsonSchema, sampleData, parentKey = '') {
  const properties = (jsonSchema?.properties || jsonSchema?.items?.properties) ?? {};
  
  Object.keys(properties).forEach((key) => {
    const fieldKey = parentKey ? `${parentKey}.${key}` : key;

    if (properties[key]['x-jsf-logic-computedAttrs']) {
      // validate computedAttrs rules for this field

      Object.values(properties[key]['x-jsf-logic-computedAttrs']).forEach((rule) => {
        checkRuleIntegrity(
          rule,
          fieldKey,
          sampleData,
          (item) =>
            `[json-schema-form] json-logic error: fieldName "${item.var}" doesn't exist in field "${fieldKey}.x-jsf-logic-computedAttrs.${key}".`
        );
      });
    }

    if (properties[key].type === 'object') {
      // recurse to validate nested object fields
      validateInlineRules(properties[key], sampleData, fieldKey);
    }

    if (properties[key].type === 'array') {
      // recurse to validate array item fields
      validateInlineRules(properties[key].items, sampleData, fieldKey);
    }
  });
}
```

I'm not gonna consider this a blocker since its a nit, but if you see what I'm doing wrong, lmk 😂

src/jsonLogic.js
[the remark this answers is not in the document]
sandrina-p: Perfect! So If I understood correctly, let's leave it as it is, and then in the next iteration consider the config to toggle it off for dev.

src/jsonLogic.js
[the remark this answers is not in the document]
sandrina-p: To me, there's no such thing as "simple regex" 😅  var all the way!

sandrina-p (approved):

src/jsonLogic.js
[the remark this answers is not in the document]
sandrina-p: It's great, thanks!

src/jsonLogic.js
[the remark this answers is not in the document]
brennj: [Internal issue created](https://linear.app/remote/issue/WEB-45)

PR #39, opened 2023-08-24
[JSON Logic] Part 5: Add general error handling

brennj:
When evaluating a rule to be validated, thats when we can get insight whether to throw or not based on the lack of existence of a field. 

Please see references of this in the example schemas provided in this PR.


src/tests/jsonLogic.test.js
```diff
@@ -98,6 +103,50 @@ describe('cross-value validations', () => {
...
+      });
+      expect(console.error).toHaveBeenCalledWith(
+        'JSON Schema invalid!',
+        Error('"child" in rule "validation_parent" does not exist as a JSON schema property.')
+      );
+    });
+
     it('Should throw when theres a missing rule', () => {
       createHeadlessForm(schemaWithMissingRule, { strictInputType: false });
       expect(console.error).toHaveBeenCalledWith(
```
dilvane, line 152: Fantastic work here with the error handling! I was wondering if we should throw an error when a rule is invalid.

Like: 

```json
"rule": {
  "++": [{ "var": "field_a" }, { "var": "field_b" }]
}
```

brennj: can't believe i didnt think of this. excellent suggestion

src/tests/jsonLogicFixtures.js
```diff
@@ -108,6 +108,77 @@ export const schemaWithMissingComputedValue = {
...
+
+export const schemaWithDeepVarThatDoesNotExist = {
+  properties: {
+    field_a: {
+      type: 'number',
+    },
+  },
+  'x-jsf-logic': {
+    validations: {
+      a_greater_than_ten: {
```
sandrina-p, line 138: not really greater than 10 anymore 😅

sandrina-p, line 138: nit-bug: not really greater than 10 anymore 😅 . To make it more generic you could just rename it to "field_a_dummy" with a dummy error message and then it's easier to copy-paste around 😅

src/tests/jsonLogic.test.js
```diff
@@ -98,6 +103,50 @@ describe('cross-value validations', () => {
...
+      );
+    });
+
+    it('A top level logic keyword will not be able to reference fieldset properties', () => {
+      createHeadlessForm(schemaWithPropertyThatDoesNotExistInThatLevelButDoesInFieldset, {
+        strictInputType: false,
+      });
+      expect(console.error).toHaveBeenCalledWith(
+        'JSON Schema invalid!',
+        Error('"child" in rule "validation_parent" does not exist as a JSON schema property.')
```
sandrina-p, line 146: question: This validation is evaluated even though it's not used anywhere in the json schema. Shall we also validate if they are linked first? Perhaps not. Related thread https://github.com/remoteoss/json-schema-form/pull/38#discussion_r1310666171. Let's close that one first before this one.

brennj:
idle thought: these merges from downstream changes are breaking my heart

src/tests/jsonLogic.test.js
brennj: note here: json-logic already does do this to a degree, but meh error handling. So I'm trying to solve it as part of the rule integrity checks (https://github.com/remoteoss/json-schema-form/pull/39/commits/d03b369e958ade8b1d3fe0455c7461fcf79ac4cc), what do you think?

src/tests/jsonLogicFixtures.js
brennj: fixed :)

src/tests/jsonLogic.test.js
brennj: I think I would rather this be a Form Building UI concern to error there. I would hope we don't run into this in production.

brennj:
This one is properly ready for another review.

johnstonbl01:
> idle thought: these merges from downstream changes are breaking my heart

Yeah... This type of process can be very brutal and feel like a lot of shuffling around / re-work. 🤗

src/tests/jsonLogic.test.js
dilvane: I think it's ok, thanks! Just one note, looking at the [json-logic-js source code](https://github.com/jwadhams/json-logic-js/blob/master/logic.js#L369), it can throw a different error message when dealing with sub rules. But Probably we don't need to deal with it for now.

dilvane (approved):

src/tests/jsonLogic.test.js
sandrina-p: Hmmmm not sure if I disagree or not. But yes, let's leave it for later when we actually use this.

sandrina-p:
> Yeah... This type of process can be very brutal and feel like a lot of shuffling around / re-work. 🤗

The 2nd most brutal thing is terrible UX of answering comments in GitHub xD

PR #43, opened 2023-08-24
[JSON Logic] Part 8: Fieldsets/Group Arrays

brennj:
LAST PART. Some of this is a little messy but I couldn't get a proper way to fix this, and by the time we've all the other parts merged in - this should be easier to clean up if we need to.

- Adds support for fieldsets having JSON logic, and individual items on a group array.
- Adds a bunch of tests for this. :D


sandrina-p:
@brennj what do you think about closing the PR and re-opening it in the future when this becomes necessary to implement?

brennj:
Closing for now!

PR #44, opened 2023-08-30
chore(fieldset): ignore values not matching the field type

sandrina-p:
chore(fieldsets): ignores values not matching the field type

### Steps to reproduce
Given a JSON Schema that expects a fieldset...

```js
{
  properties: {
    a_fieldset: {
      title: 'Fieldset title',
      description: 'Fieldset description',
      'x-jsf-presentation': {
        inputType: 'fieldset',
      },
      properties: {
        id_number: mockTextInput,
        tabs: mockNumberInput,
      },
      required: ['id_number'],
      type: 'object',
    },
  },
  required: ['a_fieldset'],
}
```

And `initialValues` that do not match the fieldset structure (object)

```js
onst result = createHeadlessForm(schemaInputTypeFieldset, {
  initialValues: {
    a_fieldset: 'foo', // should be an object instead of string
  },
});
```

The `json-schema-form` fails with an error:

> `TypeError: Cannot create property 'id_number' on string 'foo'`

### Solution

Check the `fieldset`'s value type. If it's not an object, fallback to an empty object

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.4.4-dev.20230830125207",
+  "version": "0.4.3-beta.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@remoteoss/json-schema-form",
-      "version": "0.4.4-dev.20230830125207",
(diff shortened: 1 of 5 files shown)


rylek90 (approved):
LGTM 🚀

PR #47, opened 2023-10-16
fix(select/radio): Support oneOf[].pattern validation

davelange:
Add oneOf pattern validation.
 - If the options in a oneOf contain a value, validation remains the same as it currently is.
 - If the options contain a regex pattern, use that to test the value.

How:
 - Update `radioOrSelect` yup schema
 - Add test

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -57,6 +57,7 @@ import {
   schemaInputTypeNumberWithPercentage,
   schemaForErrorMessageSpecificity,
   jsfConfigForErrorMessageSpecificity,
+  mockPatternOneOf,
 } from './helpers';
 import { mockConsole, restoreConsoleAndEnsureItWasNotCalled } from './testUtils';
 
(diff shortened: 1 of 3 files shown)


brennj:
Seems to work well, awesome work. If you can open an MR internally and ensure it passes our tests for our components, will be happy to approve this! Ping me if you need any support releasing a dev package to test

src/tests/createHeadlessForm.test.js
```diff
@@ -1442,6 +1443,45 @@ describe('createHeadlessForm', () => {
       });
     });
 
+    it('supports oneOf pattern validation', () => {
+      const result = createHeadlessForm(
+        JSONSchemaBuilder()
```
sandrina-p, line 1448: todo: `JSONSchemaBuilder` is deprecated, could you just use normal JSON?

related: #53

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.6.5-beta.0",
+  "version": "0.6.6-dev.20231026142809",
```
sandrina-p, line 3: @davelange, new version released: `0.6.6-dev.20231026142809`!

One thing, pls make sure to update your branch with master because we are already on 0.7.x

sandrina-p: @davelange could you revert the dev version? We have a script for that, it's `version_as_main`.

davelange: Done 👍

sandrina-p (approved):
Approved! I've tweaked the PR title to follow conventional commits. Releasing a new patch version in a minute.

sandrina-p:
`@remoteoss/json-schema-form@0.7.2-beta.0` released 🙌

PR #49, opened 2023-10-20
fix(select/radio): Allow numbers in oneOf

brennj:
Today, JSF will complain for the following schema:

```
{
  properties: {
    number: { type: 'number', oneOf: [{ const: 0 }, { const: 1 }, { const: 2 }] },
  },
}
```

This is because numbers aren't currently allowed to be in a oneOf list. This PR allows that schema to not be invalid.

### DIFF
--- a/src/tests/const.test.js
+++ b/src/tests/const.test.js
@@ -157,3 +157,21 @@ describe('const/default with forced values', () => {
     expect(handleValidation({ ten_only: null }).formErrors).toBeUndefined();
   });
 });
+
+describe('One of const', () => {
+  it('Should work for when the there is a list of numbers allowed', () => {
+    const { handleValidation } = createHeadlessForm(
+      {
+        properties: {
+          number: { type: 'number', oneOf: [{ const: 0 }, { const: 1 }, { const: 2 }] },
+        },
+      },
+      { strictInputType: false }
+    );
+    expect(handleValidation({}).formErrors).toEqual(undefined);
+    expect(handleValidation({ number: 3 }).formErrors).toEqual({
+      number: 'The option 3 is not valid.',
+    });
+    expect(handleValidation({ number: 2 }).formErrors).toBeUndefined();
+  });
+});
--- a/src/tests/const.test.js
+++ b/src/tests/const.test.js
@@ -177,4 +177,23 @@ describe('OneOf const', () => {
       number: 'The option "2" is not valid.',
     });
   });
+
+  it('Validates numbers or strings when type is an array with null', () => {
+    const { handleValidation } = createHeadlessForm(
+      {
+        properties: {
+          number: { type: ['number', 'null'], oneOf: [{ const: 0 }, { const: 1 }, { const: 2 }] },
+        },
+      },
+      { strictInputType: false }
+    );
+    expect(handleValidation({}).formErrors).toEqual(undefined);
+    expect(handleValidation({ number: 3 }).formErrors).toEqual({
+      number: 'The option 3 is not valid.',
+    });
+    expect(handleValidation({ number: 2 }).formErrors).toBeUndefined();
+    expect(handleValidation({ number: '2' }).formErrors).toEqual({
+      number: 'The option "2" is not valid.',
+    });
+  });
 });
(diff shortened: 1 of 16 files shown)


src/tests/const.test.js
```diff
@@ -157,3 +157,21 @@ describe('const/default with forced values', () => {
     expect(handleValidation({ ten_only: null }).formErrors).toBeUndefined();
   });
 });
+
+describe('One of const', () => {
+  it('Should work for when the there is a list of numbers allowed', () => {
```
sandrina-p, line 162:
```suggestion
describe('OneOf const', () => {
  it('Validates numbers or strings correctly', () => {
```

src/tests/const.test.js
```diff
@@ -157,3 +157,21 @@ describe('const/default with forced values', () => {
...
+          number: { type: 'number', oneOf: [{ const: 0 }, { const: 1 }, { const: 2 }] },
+        },
+      },
+      { strictInputType: false }
+    );
+    expect(handleValidation({}).formErrors).toEqual(undefined);
+    expect(handleValidation({ number: 3 }).formErrors).toEqual({
+      number: 'The option 3 is not valid.',
+    });
+    expect(handleValidation({ number: 2 }).formErrors).toBeUndefined();
```
sandrina-p, line 175: todo: Can you add a `number: '2'` to ensure strings are not valid?

sandrina-p (changes requested):

src/tests/const.test.js
brennj: good callout.... this is currently considered valid 🤦 investigating.

brennj: This baffled me a little. 

I tried to add this and it would come back as valid. However when I try to debug and run `schema.isValid("2")` it would correctly be invalid. It would be same the whole way through the `buildYupSchema` function. And yet... when we get to `handleValidation` with the lazy schema.. it considers it valid. I don't understand why. Am I missing `yup` knowledge here?

An alternative that works: using `mixed`, it seems to be more strict and won't allow strings then 🤔

src/tests/const.test.js
brennj: Added

src/tests/const.test.js
sandrina-p: I'll look into this after lunch.

sandrina-p: > I try to debug and run schema.isValid("2") it would correctly be invalid.

Are you sure? Look, both 2 and "2" are [valid on my tests](https://github.com/remoteoss/json-schema-form/assets/14869087/cca790be-ced9-4aed-ad45-a1ecc95b763b).

I didn't know about `mixed` but the way I solve this is with `strict()`, like this:

```
    number()
      .strict()
      .typeError('The value "${value}" is not a number.')
      .oneOf(...)
```

src/yupSchema.js
```diff
@@ -181,9 +188,14 @@ const getOptions = (field) => {
 const getYupSchema = ({ inputType, ...field }) => {
   const jsonType = getJsonTypeInArray(field.jsonType);
 
-  if (field.options?.length > 0) {
+  if (field.options?.length > 0 && field.jsonType !== 'number') {
+    const optionValues = getOptions(field);
+    return yupSchemas.radioOrSelectString(optionValues);
+  }
+
+  if (field.options?.length > 0 && field.jsonType === 'number') {
```
remotecom, line 196: Hi @brennj and @sandrina-p, we just test this against our form. It does not seem to handle ["number", null].

src/tests/const.test.js
paulacarneiro: I replaced `mixed()` with `number().strict()` but after further testing realized `strict` seems to also block `null`

paulacarneiro: We do need to support `null` so I changed it back to `mixed()` and added a transform

ollyd: Hey @paulacarneiro tested this out and the null option works for both radio and select, but I'm getting a validation error for radio when making a selection: `The option "1" is not valid.`. Select input worked as expected though.

paulacarneiro: Thanks for the feedback @ollyd. This is likely due to the fact that the HTML `input` element stores values as strings and we're explicitly blocking strings. For this to work with radio fields we would probably need to add some logic to the UI component to convert the value to `number`

remotecom (dismissed):

ollyd (dismissed):

paulacarneiro (approved):

sandrina-p (approved):

PR #52, opened 2023-10-23
Draft: [Experiment] Try to clear old attributes on conditionals

brennj:
TODO: fill this in and explain why its a problem

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -202,14 +202,20 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
     field.isVisible = true;
   }
 
-  const updateValues = (fieldValues) =>
+  const updateValues = (fieldValues) => {
+    ['const', 'default', 'minimum', 'maximum', 'minLength', 'maxLength'].forEach((key) => {
+      if (fieldValues[key] === undefined) {
+        field[key] = undefined;
+      }
+    });
+
     Object.entries(fieldValues).forEach(([key, value]) => {
       // some values (eg "schema") are a function, so we need to call it here
       field[key] = typeof value === 'function' ? value() : value;
 
       if (key === 'value') {
         // The value of the field should not be driven by the json-schema,
-        // unless it's a read-only field
+        // unless it's a read-only field or a forced value
         // If the readOnly property has changed, use that updated value,
         // otherwise use the start value of the property
         const readOnlyPropertyWasUpdated = typeof fieldValues.readOnly !== 'undefined';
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -202,14 +202,20 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
     field.isVisible = true;
   }
 
-  const updateValues = (fieldValues) =>
+  const updateValues = (fieldValues) => {
+    ['const', 'default', 'minimum', 'maximum', 'minLength', 'maxLength'].forEach((key) => {
+      if (fieldValues[key] === undefined) {
+        field[key] = undefined;
+      }
+    });
```
remotecom, line 210: its definitely not the way to fix this... right?

src/helpers.js
```diff
@@ -202,14 +202,20 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
     field.isVisible = true;
   }
 
-  const updateValues = (fieldValues) =>
+  const updateValues = (fieldValues) => {
+    ['const', 'default', 'minimum', 'maximum', 'minLength', 'maxLength'].forEach((key) => {
```
sandrina-p, line 206: todo: I don't need to be explicit about the downsides of having a hardcoded list of keys right? So the best we can do for now is to extract this to a function/variable and add a little but important comment about it.

p.s. It's missing `readOnly`,  `description`, and `statement`, which are very common too.

src/helpers.js
```diff
@@ -202,14 +202,20 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
...
+    });
+
     Object.entries(fieldValues).forEach(([key, value]) => {
       // some values (eg "schema") are a function, so we need to call it here
       field[key] = typeof value === 'function' ? value() : value;
 
       if (key === 'value') {
         // The value of the field should not be driven by the json-schema,
-        // unless it's a read-only field
+        // unless it's a read-only field or a forced value
```
sandrina-p, line 218: q: why do we need this line change?

src/tests/const.test.js
```diff
@@ -156,4 +156,35 @@ describe('const/default with forced values', () => {
     // Expectation: To fail with error "The only accepted value is 10."
     expect(handleValidation({ ten_only: null }).formErrors).toBeUndefined();
   });
+
+  it('Should clear the default from number when removed from condition', () => {
```
sandrina-p, line 160: todo: Can you check the tests for `schemaWithConditionalReadOnlyProperty`? Now we should be able to remove the redundant conditional `readOnly:false` (because it's the default state) and still pass.

src/helpers.js
brennj: exactly, this was thrown together, will improve! ... i think it causes other problems for descriptions, need to investigate exactly what happens

brennj:
Closing in favour of https://github.com/remoteoss/json-schema-form/pull/57

PR #53, opened 2023-10-25
chore: Mark JSONSchemaBuilder internal helper as deprecated

sandrina-p:
Why?
- It adds extra complexity to read the schema
- It's harder to copy-paste into the Playground, validators, etc

I've also refactored a few instances as example

### DIFF
--- a/src/tests/helpers.js
+++ b/src/tests/helpers.js
@@ -463,6 +463,12 @@ export const mockCheckboxInput = {
     })
     .build();
  */
+
+/**
+ * @deprecated in favor of normal JSON schema.
+ * Why? This adds extra complexity to read and to
+ * copy-paste into the Playground, validators, etc
+ * */
 export function JSONSchemaBuilder() {
   return {
     addInput: function addInput(input) {
--- a/src/tests/helpers.js
+++ b/src/tests/helpers.js
@@ -664,12 +664,12 @@ export const schemaWithoutTypes = {
   },
 };
 
-export const schemaInputTypeText = JSONSchemaBuilder()
-  .addInput({
+export const schemaInputTypeText = {
+  properties: {
     id_number: mockTextInput,
-  })
-  .setRequiredFields(['id_number'])
-  .build();
+  },
+  required: ['id_number'],
+};
 
 export const schemaInputWithStatement = JSONSchemaBuilder()
   .addInput({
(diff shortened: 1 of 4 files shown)


src/tests/helpers.js
```diff
@@ -678,83 +684,22 @@ export const schemaInputWithStatement = JSONSchemaBuilder()
...
-          description: 'This is another statement message, but more severe.',
-          inputType: 'statement',
-          severity: 'warning',
-        },
-      },
-    },
-  })
-  .build();
-
-export const schemaInputWithStatementDeprecated = JSONSchemaBuilder()
```
sandrina-p, line 712: scout: Deleted this because it's not being used (here and internally)

src/tests/helpers.js
```diff
@@ -678,83 +684,22 @@ export const schemaInputWithStatement = JSONSchemaBuilder()
...
-  })
-  .addInput({
-    a_or_b: {
-      title: 'A dropdown',
-      description: 'Some options to chose from',
-      items: {
-        enum: ['A', 'B'],
-      },
+    role: {
+      title: 'Role',
```
sandrina-p, line 688: scout: I've changed the field from select to text because:
- it was using a deprecated way of options.
- the select field was irrelevant either way for this demo.
- makes the json schema smaller, so it's easier to understand [the demo](https://github.com/remoteoss/json-schema-form/assets/14869087/e3c610a1-c21c-40d7-b7b1-f1dc7e431e7b).

remotecom:
Excellent. 🗡️

brennj (dismissed):
Looks good to me. Lets go. (will reapprove when lint fixed)

src/tests/helpers.js
```diff
@@ -463,6 +463,12 @@ export const mockCheckboxInput = {
     })
     .build();
  */
+
```
sandrina-p, line 466: The [internal pipelines are green](https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/20741). I reverted the package.json, ready for approval!

P.S. I won't do a new release after merge because this is an internal chore that doesn't affect consumers. In future release, this commit can even be deleted from the CHANGELOG.

brennj (approved):

brennj (approved):

PR #57, opened 2023-11-04
fix: Remove conditional attributes after the condition is unmatched

### LINKED ISSUE #56: Conditionals attributes are not removed after the condition is unmatched
### Problem

The following example is about `description`, but this happens with any conditional attribute (title, minimum, etc).

Given a form that asks for two fields `is_full_time` (yes/no) and `hours` (number). If yes is selected, then it shows a description warning about the maximum hours.

The JSON Schema is the following:

```json
{
  "properties": {
    "is_full_time": {
      "type": "string",
      "title": "Is full time",
      "oneOf": [{ "const": "yes", "title": "Yes" }, { "const": "no", "title": "No" }]
    },
    "hours": { "type": "number", "title": "Hours per day" }
  },
  "allOf": [
    {
      "if": {
        "properties": { "is_full_time": { "const": "yes" } },
        "required": ["is_full_time"]
      },
      "then": {
        "properties": {
          "hours": {
            "description": "We recommend no more than 8 hours."
          }
        }
      }
    }
  ]
}
```

Steps to reproduce:
1. In the [playground](https://json-schema-form.vercel.app/?path=/docs/playground--playground) add the schema above.
2. Select Yes, the description shows up.
3. Select No. The description should disappear but it remains 🐛 

image

For this scenario, a workaround could be to undo the value by set `description: ""` in the `else` statement, like this:
```
"else": {
  "properties": {
    "hours": {
      "description": ""
     }
   }
}
```

But this ain't possible for every attribute. For example, if you set `const: 10`, there's no way to "undo" it. Same for any other type of validation.

sandrina-p:
## Description

Fixes #56

How? When the fields get updated after parsing each conditional, compare the conditional attrs to the previous base field attrs. If an attr in the prev base is no longer present in the conditional, then remove it.

### DIFF
--- a/src/calculateConditionalProperties.js
+++ b/src/calculateConditionalProperties.js
@@ -79,8 +79,8 @@ function rebuildFieldset(fields, property) {
 export function calculateConditionalProperties({ fieldParams, customProperties, logic, config }) {
   /**
    * @typedef {calculateConditionalPropertiesReturn}
-   * @property {rootFieldAttrs} x - The field attributes from the first render (root)
-   * @property {newAttributes} y - Attributes from the matched conditional
+   * @property {rootFieldAttrs} - The field attributes from the first render (root)
+   * @property {newAttributes} - Attributes from the matched conditional
    */
   /**
    * Runs dynamic property calculation on a field based on a conditional that has been calculated
--- a/src/calculateConditionalProperties.js
+++ b/src/calculateConditionalProperties.js
@@ -69,7 +69,7 @@ function rebuildFieldset(fields, property) {
  * schema dependencies, and conditional logic.
  *
  * @param {Object} params - Parameters
- * @param {Object} params.fieldParams - Current field parameters
+ * @param {Object} params.fieldParams - The field attributes from the first render (root)
  * @param {Object} params.customProperties - Custom field properties from schema
  * @param {Object} params.logic - JSON-logic
  * @param {Object} params.config - Form configuration
@@ -78,6 +78,11 @@ function rebuildFieldset(fields, property) {
  */
 export function calculateConditionalProperties({ fieldParams, customProperties, logic, config }) {
   /**
+   * @typedef {calculateConditionalPropertiesReturn}
+   * @property {rootFieldAttrs} x - The field attributes from the first render (root)
+   * @property {newAttributes} y - Attributes from the matched conditional
+   */
+  /**
    * Runs dynamic property calculation on a field based on a conditional that has been calculated
    *
    * @param {Object} params - Parameters
@@ -85,11 +90,11 @@ export function calculateConditionalProperties({ fieldParams, customProperties, 
    * @param {Object} params.conditionBranch - Condition branch
    * @param {Object} params.formValues - Current form values
    *
-   * @returns {Object} Updated field parameters
+   * @returns {calculateConditionalPropertiesReturn}
    */
   return ({ isRequired, conditionBranch, formValues }) => {
     // Check if the current field is conditionally declared in the schema
-
+    // console.log('::calc (closure original)', fieldParams.description);
     const conditionalProperty = conditionBranch?.properties?.[fieldParams.name];
 
     if (conditionalProperty) {
@@ -142,20 +147,26 @@ export function calculateConditionalProperties({ fieldParams, customProperties, 
         ),
       };
 
-      return omit(merge(base, presentation, newFieldParams), ['inputType']);
+      return {
+        rootFieldAttrs: fieldParams,
+        newAttributes: omit(merge(base, presentation, newFieldParams), ['inputType']),
+      };
     }
 
     // If field is not conditionally declared it should be visible if it's required
     const isVisible = isRequired;
 
     return {
-      isVisible,
-      required: isRequired,
-      schema: buildYupSchema({
-        ...fieldParams,
-        ...extractParametersFromNode(conditionBranch),
+      rootFieldAttrs: fieldParams,
+      newAttributes: {
+        isVisible,
         required: isRequired,
-      }),
+        schema: buildYupSchema({
+          ...fieldParams,
+          ...extractParametersFromNode(conditionBranch),
+          required: isRequired,
+        }),
+      },
     };
   };
 }
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -14,43 +14,34 @@ import { buildCompleteYupSchema, buildYupSchema } from './yupSchema';
 
 /**
  * List of attributes for each JSF' field
- * that do not exist in JSON Schema directly.
- * [*1] These are not covered by any unit-test yet
+ * that are added dynamically and not exist in JSON Schema directly.
  */
-const fieldAttrsFromJsf = [
-  'name', // driven from json schema properties' name
-  'label', // driven from schema field title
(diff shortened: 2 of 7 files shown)


src/helpers.js
```diff
@@ -12,6 +12,44 @@ import { processJSONLogicNode } from './jsonLogic';
...
+/**
+ *
+ * @param {*} finalField
+ * @param {*} branchAttrs
+ */
+function removeStaleAttrsFromField(field, latestAttrs) {
+  Object.keys(field).forEach((key) => {
+    if (
+      latestAttrs[key] === undefined &&
+      fieldJsfAttrsObj[key] === undefined // ignore these because they are internal
```
sandrina-p, line 45: note: In the original PR (https://github.com/remoteoss/json-schema-form/pull/52) it was created a list of attrs that can be removed. In this PR, I think it's safer to do it the other way around: create a list of fields that cannot be removed, which is basically every internal computed JSF attr (`fieldJsfAttrsObj`).

I still don't think this is the ideal solution and the real problem is somewhere earlier in the data flow.

But based on the urgency to unblock other features/consumers and the strong unit-tests, I'm comfortable with this solution.

src/tests/conditions.test.js
```diff
@@ -56,3 +51,308 @@ it('Should allow check of a nested property in a conditional', () => {
...
+    // Changing to "No", it applies the base value again.
+    expect(handleValidation({ is_full_time: 'no' }).formErrors).toBeUndefined();
+    expect(fields[1].anything).toBe('info');
+
+    // Changing back to "Yes", it sets the attribute again
+    expect(handleValidation({ is_full_time: 'yes' }).formErrors).toBeUndefined();
+    expect(fields[1].anything).toBe('danger');
+  });
+
+  it('Keeps existing attributes in matches that dont change the attr', () => {
```
sandrina-p, line 249: bug: Our internal Storybook tests spot this scenario as failing. I'm looking into a fix right now.

sandrina-p: I think I cracked it, ready for review!

Change summary: We cannot compare the `conditionalAttrs` to the current`field` alone, we also need to compare it to the original field (the ones in the root).

src/helpers.js
```diff
@@ -202,19 +237,19 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
     field.isVisible = true;
   }
 
-  const updateValues = (fieldValues) =>
-    Object.entries(fieldValues).forEach(([key, value]) => {
-      // some values (eg "schema") are a function, so we need to call it here
+  const updateAttributes = (fieldAttrs) => {
```
sandrina-p, line 240: nit: I renamed it from `updateValues` because I thought it was about the field value (eg has_pet: "yes")

src/helpers.js
```diff
@@ -226,36 +261,38 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
...
   if (field.calculateConditionalProperties) {
-    const newFieldValues = field.calculateConditionalProperties({
+    const { rootFieldAttrs, newAttributes } = field.calculateConditionalProperties({
       isRequired: fieldIsRequired,
       conditionBranch: node,
       formValues,
     });
-    updateValues(newFieldValues);
+    updateAttributes(newAttributes, rootFieldAttrs);
+    removeConditionalStaleAttributes(field, newAttributes, rootFieldAttrs);
```
sandrina-p, line 286: note: In the first commit the `removeConditionalStaleAttributes` was inside `updateAttributes` so it would be called for every dynamic scenario (conditionals, custom validations, and computed attrs). But tha it caused other tests to fail, and I couldn't understand why in a short time.

As this bug is specifically focused on "leaked" attributes in conditionals, I decided to focus the solution only on that for now. I never saw bugs reported from the other dynamic scenarios, so... as they say "[if it ain't broke, don't fix it](https://en.wikipedia.org/wiki/wikt:if_it_ain%27t_broke,_don%27t_fix_it)." 

A conscious trade-off between between speed and excellence.

src/helpers.js
```diff
@@ -226,36 +261,38 @@ function updateField(field, requiredFields, node, formValues, logic, config) {
...
   // If field has a calculateConditionalProperties closure, run it and update the field properties
   if (field.calculateConditionalProperties) {
-    const newFieldValues = field.calculateConditionalProperties({
+    const { rootFieldAttrs, newAttributes } = field.calculateConditionalProperties({
       isRequired: fieldIsRequired,
       conditionBranch: node,
       formValues,
     });
-    updateValues(newFieldValues);
+    updateAttributes(newAttributes, rootFieldAttrs);
```
remotecom, line 285: You're passing in `rootFieldAttrs` but as far as I can see, it's not used, right?

brennj, line 285: You're passing in `rootFieldAttrs` but as far as I can tell, its not used right?

brennj (commented):
Just one small thing needs to be changed and its only a lint error. Otherwise, tests are fantastic and I'm so happy we have this now.

src/helpers.js
sandrina-p: Ups, a tail from previous commits, removed!

src/helpers.js
```diff
@@ -12,6 +12,41 @@ import { processJSONLogicNode } from './jsonLogic';
...
+ *
+ * @param {Object} field - Current field attributes
+ * @param {Object} conditionalAttrs - Attributes from the matched conditional
+ * @param {Object} rootAttrs - Original field attributes from the root.
+ */
+function removeConditionalStaleAttributes(field, conditionalAttrs, rootAttrs) {
+  Object.keys(field).forEach((key) => {
+    if (
+      conditionalAttrs[key] === undefined &&
+      rootAttrs[key] === undefined && // Don't remove attrs that were declared in the root field.
```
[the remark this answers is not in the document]
sandrina-p: Yes, also the `dynamicInternalJsfAttrsObj` (which is the opposite of your initial solution).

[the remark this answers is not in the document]
sandrina-p: > [@brennj] Just one small thing needs to be changed and its only a lint error. Otherwise, tests are fantastic and I'm so happy we have this now.

Cool! I'm gonna wait for the internal pipelines to be green before merging this one.

dilvane (commented):
This is a fantastic fix! Thanks for this, until now, we were using the "workaround" to override the attribute like you showed in the playground, but with the "const" the need for this fix is more visible!

src/calculateConditionalProperties.js
```diff
@@ -69,14 +69,19 @@ function rebuildFieldset(fields, property) {
...
  * @param {Object} params.config - Form configuration
  *
  * @returns {Function} A function that calculates conditional properties
  */
 export function calculateConditionalProperties({ fieldParams, customProperties, logic, config }) {
+  /**
+   * @typedef {calculateConditionalPropertiesReturn}
+   * @property {rootFieldAttrs} x - The field attributes from the first render (root)
+   * @property {newAttributes} y - Attributes from the matched conditional
+   */
```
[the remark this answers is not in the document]
sandrina-p: Ups, fixed! I also removed the dev version for final approval

brennj (approved):

sandrina-p:
`@remoteoss/json-schema-form@0.7.3-beta.0` released 📦

PR #54, opened 2023-10-27
fix(select): support string type

joeynimu:
**Description** 

This MR adds support for select field types with a string type option. 

**Usecase**

When working with `select` fields, at times you might want the user to either pick an existing option or create an option on the fly. 

**Usage**

```
{
...
  "department": {
      "title": "Department",
      "type": "string",
      "description": "Select a department or create one",
      "x-jsf-presentation": {
        "inputType": "select"
      },
      "oneOf": [
        { "const": "engineering_slug", "title": "Engineering" },
        { "const": "marketing_slug", "title": "Marketing" },
        { "const": "sales_slug", "title": "Sales" },
        { "type": "string", "title": "{Create another}" }
      ]
    },
 ...
}
```

PR #62, opened 2023-12-19
fix(select): Remove creatable logic

remotecom:
This PR removes the logic added in https://github.com/remoteoss/json-schema-form/pull/54 for we will no longer be using this approach.

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -632,8 +632,8 @@ describe('createHeadlessForm', () => {
       });
     });
 
-    it('supports "select" field type with string option', () => {
-      const { fields, handleValidation } = createHeadlessForm(schemaInputTypeSelectString);
+    it('supports "select" field type with creatable option', () => {
+      const { fields } = createHeadlessForm(schemaInputTypeSelectString);
       const fieldSelect = fields[0];
       expect(fieldSelect).toMatchObject({
         name: 'browsers',
@@ -657,12 +657,6 @@ describe('createHeadlessForm', () => {
       });
 
       expect(fieldSelect).not.toHaveProperty('multiple');
-
-      assertOptionsAllowed({
-        handleValidation,
-        fieldName: 'browsers',
-        validOptions: ['chr', 'ff', 'ie'],
-      });
     });
 
     it('supports "select" field type with multiple options @deprecated', () => {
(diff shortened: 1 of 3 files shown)


src/tests/createHeadlessForm.test.js
```diff
@@ -639,8 +632,8 @@ describe('createHeadlessForm', () => {
       });
     });
 
-    it('supports "select" field type with string option', () => {
-      const { fields, handleValidation } = createHeadlessForm(schemaInputTypeSelectString);
+    it('supports "select" field type with creatable option', () => {
```
sandrina-p, line 635: question: Why keeping this test if this is no longer a "creatable" select?

src/tests/helpers.js
```diff
@@ -240,7 +240,11 @@ export const mockSelectInputSolo = {
 
 export const mockSelectInputSoloCreatable = {
   ...mockSelectInputSolo,
-  oneOf: [...mockSelectInputSolo.oneOf, { type: 'string', title: '{Create another}' }],
+  creatable_on: 'browser_name',
```
sandrina-p, line 243: question: I don't understand, why do we need this new keyword in json-schema-form? How does it influence the mechanisms of `json-schema-form` itself?

If I'm understanding correctly, this MR is just about removing the select support for `type: "string"` because we no longer need it internally. Or am I missing something?

sandrina-p (changes requested):

src/tests/helpers.js
remotecom: @sandrina-p sorry, makes total sense. Let me update it and tag you as a reviewer

sandrina-p: @joeynimu you are logged with Remotecom account, but you should be using your own account :)

joeynimu: yikes!

sandrina-p: Okay the code is coode now! The missing part is to update the PR title and description to reflect what was really done. Once you do that, I'll approve the PR :)

joeynimu: @sandrina-p took me a while to come back to this, I can blame PTO + other priorities but I have updated the title and description. Let me know if this works.

sandrina-p: Thanks! The correct naming though is `fix(select): Remove creatable logic". The select makes it explicit this is for select fields. "Remove" make it explicit that we removed a unwanted feature, "clean up" sound more like tyding/refactor.

Approved :)

sandrina-p: To make it clear, this can still be useful in upcoming cases, but as it's not needed right now, better to delete it. We can always add it back later when needed.

sandrina-p (approved):

src/tests/helpers.js
joeynimu: @sandrina-p noted and updated ✅ thanks for the review

PR #65, opened 2024-02-07
fix(conditionals): Certain conditions were failing due to lack of context

brennj:
Consider the following schema: 

```json
{
  "allOf": [
    {
      "if": {
        "properties": {
          "salary": {
            "maximum": 119999
          }
        },
        "required": ["salary"]
      },
      "then": {
        "required": ["reason"]
      },
      "else": {
        "properties": {
          "reason": false
        }
      }
    }
  ],
  "properties": {
    "salary": {
      "type": "number",
      "x-jsf-presentation": {
        "inputType": "money"
      }
    },
    "reason": {
      "oneOf": [
        {
          "const": "reason_one"
        },
        {
          "const": "reason_two"
        }
      ],
      "type": "string"
    }
  },
  "required": ["salary"],
  "type": "object"
}
```

With the presence of the `x-jsf-presentation` in the `salary` field, it is causing a `Error: Cannot read properties of undefined (reading 'isValidSync')` error to occur.

To fix this in this PR, we must ensure all existing properties are being passed down to further conditionals being evaluated.

### DIFF
--- a/src/checkIfConditionMatches.js
+++ b/src/checkIfConditionMatches.js
@@ -59,10 +59,8 @@ export function checkIfConditionMatchesProperties(node, formValues, formFields, 
 
     return validateFieldSchema(
       {
-        options: field.options,
-        // @TODO/CODE SMELL. We are passing the property (raw field), but buildYupSchema() expected the output field.
-        ...{ ...field, ...currentProperty },
-        inputType: field.inputType,
+        ...field,
+        ...currentProperty,
         required: true,
       },
       value
--- a/src/checkIfConditionMatches.js
+++ b/src/checkIfConditionMatches.js
@@ -61,7 +61,7 @@ export function checkIfConditionMatchesProperties(node, formValues, formFields, 
       {
         options: field.options,
         // @TODO/CODE SMELL. We are passing the property (raw field), but buildYupSchema() expected the output field.
-        ...currentProperty,
+        ...{ ...field, ...currentProperty },
         inputType: field.inputType,
         required: true,
       },
--- a/src/tests/conditions.test.js
+++ b/src/tests/conditions.test.js
@@ -439,11 +439,10 @@ describe('Conditional with a minimum value check', () => {
             required: ['salary'],
           },
           then: {
-            required: ['employee_knows', 'reason'],
+            required: ['reason'],
           },
           else: {
             properties: {
-              employee_knows: false,
               reason: false,
             },
           },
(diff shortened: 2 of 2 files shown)


vanessaloviton (approved):

brennj:
@sandrina-p can i get you to approve these changes when you have a chance please? :)

src/checkIfConditionMatches.js
```diff
@@ -59,10 +59,8 @@ export function checkIfConditionMatchesProperties(node, formValues, formFields,
 
     return validateFieldSchema(
```
sandrina-p, line 60: > [@brennj] @sandrina-p can i get you to approve these changes when you have a chance please? :)

Do you need it today or can it be Monday morning?

brennj: Whenever works ;)

src/checkIfConditionMatches.js
```diff
@@ -59,10 +59,8 @@ export function checkIfConditionMatchesProperties(node, formValues, formFields,
...
       {
-        options: field.options,
-        // @TODO/CODE SMELL. We are passing the property (raw field), but buildYupSchema() expected the output field.
+        ...field,
         ...currentProperty,
-        inputType: field.inputType,
```
sandrina-p, line 65: though: I wonder why we never passed the entire `...fied` before. Perhaps to avoid some other bug? 🤔 

Pipelines are green, both here and the [internally repo](https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/23636), so I'm trusting this won't break something else.

sandrina-p (approved):

src/tests/conditions.test.js
```diff
@@ -423,3 +423,59 @@ describe('Conditional attributes updated', () => {
     expect(fields[1].visibilityCondition).toEqual(expect.any(Function));
   });
 });
+
```
sandrina-p, line 426: nit: A better PR title (commit) name:

```diff
- fix(conditionals): Certain conditions were failing due to lack of context #65
- fix(conditionals): Add missing field context to conditions #65
```

sandrina-p: But to be honest I'm not sure 🤔 We don't have any guidelines either. Thoughts?

sandrina-p: Can be a mix of both. Problem and solution.

```
fix(conditionals): Certain conditions were failing. Add missing field context to evaluation
```

src/checkIfConditionMatches.js
brennj: i apparently implemented some of this before, and I'm pretty sure I was leaning on the conservative side not wanting to pass down everything unless necessary, and this is how it manifested ... 6+ months later.

src/tests/conditions.test.js
brennj: Love it, will use for merge commit.

sandrina-p: but up to you!

sandrina-p: (oh github failed on me for the updates, perfect then)

PR #66, opened 2024-02-09
Rename `value` -> `forcedValue`

brennj:
We added the custom attribute value, which has some problems:

Bug: We set it when const and default exist, but we do not check if both are equal (they must).

API: The word itself clashes with native HTML attribute (value) which can create confusing in the UI integration, cause unexpected bugs with a field stuck in a value.

This PR names it from `value` to `forcedValue`


paulacarneiro:
Tested this with different field types and it's working well 🙌🏼

paulacarneiro (approved):

PR #67, opened 2024-02-13
fix(helpers): getFieldOptions - return empty array if oneOf missing in radio

rylek90:
Previously a valid schema with `inputType: radio` property, but without `oneOf` resulted in options returned as `null` which broke integration in the product. This addresses that issue.

https://remote-com.slack.com/archives/C02HTN0LY02/p1707816400971139?thread_ts=1706863325.161439&cid=C02HTN0LY02

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.8.1-beta.0",
+  "version": "0.8.2-dev.20240213104942",
   "lockfileVersion": 2,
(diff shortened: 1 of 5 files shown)


brennj:
Looks good to me, you just need to verify it works internally aganist our FE following the instructions [here](https://github.com/remoteoss/json-schema-form/blob/main/CONTRIBUTING.md#public-release) to release a dev package. Once we see green in the CI, lets get this in.

brennj (approved):
Approving based on passing tests internally.

PR #73, opened 2024-04-25
feat: support mapping external types to internal input types

juicycleff:
### Problem
`Json Schema Form` supports internal types and treats them quite differently eg. `group-array`. If a system requires multiple kind of components to be treated internally as group-array, this would be impossible.

### Solution
This PR adds support for mapping external input types as group array eg.

```typescript
const headlessFormOutput = createHeadlessForm(schema, {
    ...,
    presentationMapping: {
      groupArray: ["array", "conditions"],
    }
  })
```

### DIFF
--- a/src/createHeadlessForm.js
+++ b/src/createHeadlessForm.js
@@ -57,12 +57,18 @@ import { buildYupSchema } from './yupSchema';
  */
 
 /**
+ * @typedef {Object} PresentationMapping
+ * @property {Array} [groupArray] - map external input types as group-array
+ */
+
+/**
  * @typedef {Object} JsfConfig
  * @property {Object} [config.initialValues] - Initial values to evaluate the form against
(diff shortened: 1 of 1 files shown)


sandrina-p:
Hi @juicycleff! Forgive us for our delay with an answer. Could you better explain what you are trying to solve here (give us an example)? Perhaps we can propose a better solution that does not require adding a new config `presentationMapping`.

juicycleff:
Hi @sandrina-p yes you are spot on and in a separate version for nested of any type. I will finish up. Thanks.

sandrina-p:
Hi @juicycleff, checking in when you plan to have this done :)

sandrina-p:
Hi @juicycleff, could you answer here? If we don't get an answer until the end of the year, we'll close the PR :)

sandrina-p:
I created an issue https://github.com/remoteoss/json-schema-form/issues/106 for visibility

PR #74, opened 2024-05-19
chore(Text field docs): Change naming of ID number property

ollyd:
Changes the Text field property in our internal docs and JSON Schema Form docs to have slightly less confusing naming. Rather than ID Number and digits which sounds like numbers, it's now Username and characters.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.9.2-dev.20240518112325",
+  "version": "0.10.0-beta.0",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.9.1-beta.0",
+  "version": "0.9.2-dev.20240518112325",
```
sandrina-p, line 3: Great, thanks! Do you have a MR in our internal codebase to ensure the pipelines are still green? Can you link it here?

ollyd: Yep, corresponding MR is [here](https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/26523).

sandrina-p: Perfect, then yes, let's go with this one. Revert the version with `npm run version_as_main` and we can merge this

sandrina-p (dismissed):

sandrina-p (approved):

PR #78, opened 2024-07-11
feat(modify): Allow customize fields order

sandrina-p:
Follow-up of https://github.com/remoteoss/json-schema-form/pull/71, scoped on customizing the order of fields.

This MR can only be merged after #83 

[Linear issue (internal)](https://linear.app/remote/issue/PBYR-1363)

### New specs

Imagine the schema below with 4 fields:

```js
{
  properties: {
    /* does not matter */
  },
  'x-jsf-order': ['field_a', 'field_b', 'field_c', 'field_d'],
};

```

To reorder the fields, pass the given config:

```js
const result = modify(schema, {
  orderRoot: ['field_d', 'field_c', 'field_a', 'field_b',];
});
```

The new order will be: `['field_d', 'field_c', 'field_a', 'field_b',];`.

### As callback

You also have access to the original order using a callback:

```js
const result = modify(baseExample, {
  orderRoot: (original) => original.reverse(),
});
```

The new order will be  `['field_d', 'field_c', 'field_b', 'field_a']`. 

### Missing fields 

By default, any missing field will the automatically appended to the final order. 

In this example, the missing fields field_a and field_c are still included in the final order.

```js
const { schema, warnings }  = modify(schema, {
  orderRoot: ['field_b', 'field_d'];
});

console.log(warning)
{
  type: 'ODER_MISSING_FIELDS',
  message: 'Some fields got forgotten in the new order. They were automatically appended to the end: field_a, field_c',
},

```
---

#### Features NOT included:
- `orderRoot` supporting nested fields (fieldsets). (check tests for workaround)

### DIFF
--- a/src/modify.js
+++ b/src/modify.js
@@ -125,7 +125,7 @@ export function modify(originalSchema, config) {
   const resultRewrite = rewriteFields(schema, config.fields);
   rewriteAllFields(schema, config.allFields);
 
-  const resultReorder = reorderFields(schema, config.rootOrder);
+  const resultReorder = reorderFields(schema, config.orderRoot);
 
   if (!config.muteLogging) {
     console.warn(
--- a/src/modify.js
+++ b/src/modify.js
@@ -1,3 +1,4 @@
+import difference from 'lodash/difference';
 import get from 'lodash/get';
 import mergeWith from 'lodash/mergeWith';
 
@@ -19,7 +20,11 @@ function mergeReplaceArray(_, newVal) {
 }
 
 function standardizeAttrs(attrs) {
-  const { errorMessage, properties, ...rest } = attrs;
+  const {
+    errorMessage, // this is the key that will be renamed
+    properties, // destructured because of recursive call afterwards
+    ...rest
+  } = attrs;
 
   return {
     ...rest,
@@ -62,6 +67,10 @@ function rewriteFields(schema, fieldsConfig) {
       const result = rewriteFields(get(schema.properties, fieldPath), fieldChanges.properties);
       warnings.push(result.warnings);
     }
+
+    if (fieldChanges.order) {
+      reorderFields(fieldAttrs, fieldChanges.order);
+    }
   });
 
   return { warnings: warnings.flat() };
@@ -92,12 +101,29 @@ function rewriteAllFields(schema, configCallback, context) {
   });
 }
 
+function reorderFields(schema, orderCallback) {
+  if (!orderCallback) return null;
+
+  const originalOrder = schema['x-jsf-order'];
+  const orderConfig = orderCallback(originalOrder);
+  const remaining = difference(originalOrder, orderConfig.order);
+
+  const finalOrder =
+    orderConfig.rest === 'end'
+      ? [...orderConfig.order, ...remaining]
+      : [...remaining, ...orderConfig.order];
+
+  schema['x-jsf-order'] = finalOrder;
+}
+
 export function modify(originalSchema, config) {
   const schema = JSON.parse(JSON.stringify(originalSchema));
   // All these functions mutate "schema" that's why we create a copy above
 
   const resultRewrite = rewriteFields(schema, config.fields);
   rewriteAllFields(schema, config.allFields);
+
+  reorderFields(schema, config.order);
 
   if (!config.muteLogging) {
     console.warn(
--- a/src/modify.js
+++ b/src/modify.js
@@ -4,7 +4,7 @@ import mergeWith from 'lodash/mergeWith';
 
 const WARNING_TYPES = {
   FIELD_TO_CHANGE_NOT_FOUND: 'FIELD_TO_CHANGE_NOT_FOUND',
-  ODER_MISSING_FIELDS: 'ODER_MISSING_FIELDS',
+  ORDER_MISSING_FIELDS: 'ORDER_MISSING_FIELDS',
 };
 /**
  *
@@ -104,7 +104,7 @@ function reorderFields(schema, configOrder) {
 
   if (remaining.length > 0) {
     warnings.push({
-      type: WARNING_TYPES.ODER_MISSING_FIELDS,
+      type: WARNING_TYPES.ORDER_MISSING_FIELDS,
       message: `Some fields got forgotten in the new order. They were automatically appended: ${remaining.join(
         ', '
       )}`,
--- a/src/modify.js
+++ b/src/modify.js
@@ -109,9 +109,9 @@ function reorderFields(schema, orderCallback) {
   const remaining = difference(originalOrder, orderConfig.order);
 
   const finalOrder =
-    orderConfig.rest === 'end'
-      ? [...orderConfig.order, ...remaining]
-      : [...remaining, ...orderConfig.order];
+    orderConfig.rest === 'start'
+      ? [...remaining, ...orderConfig.order]
+      : [...orderConfig.order, ...remaining];
 
   schema['x-jsf-order'] = finalOrder;
 }
--- a/src/modify.js
+++ b/src/modify.js
@@ -21,11 +21,7 @@ function mergeReplaceArray(_, newVal) {
 }
 
 function standardizeAttrs(attrs) {
-  const {
-    errorMessage, // this is the key that will be renamed
-    properties, // destructured because of recursive call afterwards
-    ...rest
-  } = attrs;
+  const { errorMessage, properties, ...rest } = attrs;
 
   return {
     ...rest,
@@ -105,10 +101,11 @@ function reorderFields(schema, configOrder) {
   const originalOrder = schema['x-jsf-order'];
   const orderConfig = typeof configOrder === 'function' ? configOrder(originalOrder) : configOrder;
   const remaining = difference(originalOrder, orderConfig);
+
   if (remaining.length > 0) {
     warnings.push({
       type: WARNING_TYPES.ODER_MISSING_FIELDS,
-      message: `Some fields got forgotten in the new order. They were automatically appended to the end: ${remaining.join(
+      message: `Some fields got forgotten in the new order. They were automatically appended: ${remaining.join(
         ', '
       )}`,
     });
--- a/src/modify.js
+++ b/src/modify.js
@@ -108,12 +108,7 @@ function reorderFields(schema, orderCallback) {
   const orderConfig = orderCallback(originalOrder);
   const remaining = difference(originalOrder, orderConfig.order);
 
-  const finalOrder =
-    orderConfig.rest === 'start'
-      ? [...remaining, ...orderConfig.order]
-      : [...orderConfig.order, ...remaining];
-
-  schema['x-jsf-order'] = finalOrder;
+  schema['x-jsf-order'] = [...orderConfig.order, ...remaining];
 }
 
 export function modify(originalSchema, config) {
--- a/src/modify.js
+++ b/src/modify.js
@@ -98,7 +98,7 @@ function reorderFields(schema, configOrder) {
   if (!configOrder) return { warnings: null };
 
   const warnings = [];
-  const originalOrder = schema['x-jsf-order'];
+  const originalOrder = schema['x-jsf-order'] || [];
   const orderConfig = typeof configOrder === 'function' ? configOrder(originalOrder) : configOrder;
   const remaining = difference(originalOrder, orderConfig);
 
--- a/src/modify.js
+++ b/src/modify.js
@@ -4,6 +4,7 @@ import mergeWith from 'lodash/mergeWith';
 
 const WARNING_TYPES = {
   FIELD_TO_CHANGE_NOT_FOUND: 'FIELD_TO_CHANGE_NOT_FOUND',
+  ODER_MISSING_FIELDS: 'ODER_MISSING_FIELDS',
 };
 /**
  *
@@ -106,7 +107,7 @@ function reorderFields(schema, configOrder) {
   const remaining = difference(originalOrder, orderConfig);
   if (remaining.length > 0) {
     warnings.push({
-      type: 'reorderFields',
+      type: WARNING_TYPES.ODER_MISSING_FIELDS,
       message: `Some fields got forgotten in the new order. They were automatically appended to the end: ${remaining.join(
         ', '
       )}`,
--- a/src/modify.js
+++ b/src/modify.js
@@ -67,10 +67,6 @@ function rewriteFields(schema, fieldsConfig) {
       const result = rewriteFields(get(schema.properties, fieldPath), fieldChanges.properties);
       warnings.push(result.warnings);
     }
-
-    if (fieldChanges.order) {
-      reorderFields(fieldAttrs, fieldChanges.order);
-    }
   });
 
   return { warnings: warnings.flat() };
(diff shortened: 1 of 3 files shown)


src/tests/modify.test.js
```diff
@@ -298,3 +306,136 @@ describe('modify() - basic mutations', () => {
...
+            },
+          },
+        };
+      },
+    });
+    */
+    expect(resultA).toMatchObject({
+      properties: {
+        address: {
+          'x-jsf-order': ['city', 'zipcode', 'first_line'],
```
sandrina-p, line 411: note/thought: I had two approaches for this API:

The one committed is what I call approach A, then the approach B is:

```js
    const resultB = modify(baseExample, {
      order: () => {
        return {
          order: (original) => {
            return original.reverse(); // ['age', 'address']
          },
          fields: {
            address: {
              order: (original) => {
                return {
                  order: original.reverse(), // ['city', 'zipcode', 'first_line']
                };
              },
            },
          },
        };
      },
    });

```

In other words:

- Approach A: The API based on field scope. The downside is that the root `order:` is just for the root, nothing else. Is is misleading?
- Approach B: The API is based on order scope. The biggest downside is that it's harder to isolate customizations per fieldset. Let's imagine you have multiple customizations to the `address` fieldset, you'd be forced to declare them in two places...

What do you think?

src/modify.js
```diff
@@ -66,12 +75,31 @@ function rewriteAllFields(schema, configCallback, context) {
...
 
+function reorderFields(schema, orderCallback) {
+  if (!orderCallback) return null;
+
+  const originalOrder = schema['x-jsf-order'];
+  const orderConfig = orderCallback(originalOrder);
+  const remaining = difference(originalOrder, orderConfig.order);
+
+  const finalOrder =
+    orderConfig.rest === 'start'
```
eng-almeida, line 86: It took me some time to understand what `rest` means. Have you thought about using other parameter name? Suggestion: `remainingFieldsPosition`?

src/tests/modify.test.js
```diff
@@ -298,3 +306,112 @@ describe('modify() - basic mutations', () => {
...
+      properties: {
+        /* does not matter */
+      },
+      'x-jsf-order': ['field_a', 'field_b', 'field_c', 'field_d'],
+    };
+    const result = modify(baseExample, {
+      order: () => {
+        return {
+          order: ['field_c', 'field_a', 'field_b'],
+          rest: 'start',
```
eng-almeida, line 344: A different naming approach that I'm thinking:

```
return {
  fields: ['field_c', 'field_a', 'field_b'],
  rest: 'start',
}
```

I usually associate order with desc/asc, so we maybe this naming would be more explicit?

src/tests/modify.test.js
eng-almeida: I think everything related to order, should be under order property. So, I agree that option A is a bit misleading. Based on my comment above, I believe option B could work better but we some changes (suggestions)

```
return {
  fields: ['age', ['address', (original) => original.reverse() ]]
  rest: 'start',
}
```

Let me know your thoughts!

eng-almeida (commented):
@sandrina-p I have only a few questions/suggestions from the architecture point of view. Let me know your thoughts!

src/modify.js
sandrina-p: Can we find a middle name that is not so long, but not so short? What about `leftovers` or `remaining`

src/tests/modify.test.js
sandrina-p: Oh good suggestion, I'll apply it!

src/tests/modify.test.js
sandrina-p: Uhhh! I can tell you got that inspiration from Storybook sorting config eheh

But what about this downside:

> The biggest downside is that it's harder to isolate customizations per fieldset. Let's imagine you have multiple customizations to the address fieldset and its nested fields... you'd be forced to declare them in two places...

And another problem: How would you override 'address' without touching/redefining the root order?

sandrina-p: Regarding the API... here's another reason why I prefer approach B. To follow the same mindset of native JSON Schemas:
- in the root we only declare the root required fields, and the root conditionals
- when you need nested required fields or conditionals, you do it it in their own "fieldset" scope.

I feel like it's easier to isolate and customize if we do the same here? 🤔

sandrina-p: Btw, check the comments in the `create` MR.  https://github.com/remoteoss/json-schema-form/pull/80. It's important the API is consistent across all methods (order, pick, create).

src/modify.js
eng-almeida: Tbh I would rather use a more descriptive name. Putting myself in a developer's shoes, none of these field names seem intuitive. That's why I suggest a longer name.

sandrina-p: We had a sync chat, and we concluded to remove `rest` attr all together. If the devs wants to customize this they can always use the callback `(originalOrder) => {}` to apply any desired manipulation

src/tests/modify.test.js
sandrina-p: Done in the latest commit

src/tests/modify.test.js
eng-almeida: > in the root we only declare the root required fields, and the root conditionals

Even though I'm not a big a fan of JSON Schema readability, I'm sold to this point, so option B seems like the most reasonable approach 👍

sandrina-p: @danielcardoso5 we need you here to help break the tie on the API x)

sandrina-p: > Even though I'm not a big a fan of JSON Schema readability

Make us two! 🤷 Consistency has a big weight here.


The only downside I see with Option A is when we face a JSON Schema like this:

```js
{
  properties: {
    address: {
      "title": "My address",
      properties: { ... },
      'x-jsf-order': ['first_line', 'zipcode', 'city'],
      order: "priority",
    },
    age: { ... },
  },
  'x-jsf-order': ['address', 'age'],
};

The `address.order: priority" could be used in a scenario for something else totally different and would clash with our modify Config API... He don't have anything like that nowadays, but a random case in the OSS world might bring it. 

If we haver face it and are forced to adapt, in those cases we could change our API to:

```js
modify(baseExample, {
  fields: {
    address: {
      'x-jsf-order': (original) => {
        return {
          fields: original.reverse()
        };
      },
    },
  },
});
```

sandrina-p: I have to leave earlier today, but I can implement this fallback 'x-jsf-order' on Monday if you agree.

src/modify.js
```diff
@@ -66,12 +75,26 @@ function rewriteAllFields(schema, configCallback, context) {
   });
 }
 
+function reorderFields(schema, configOrder) {
+  if (!configOrder) return null;
+
+  const originalOrder = schema['x-jsf-order'];
+  const orderConfig = typeof configOrder === 'function' ? configOrder(originalOrder) : configOrder;
+  const remaining = difference(originalOrder, orderConfig.fields);
```
danielcardoso5, line 83: This is also a good place to put an `onWarn` call to flag to devs that a field that was previously ordered was not passed in the new order config

```suggestion
  const nonOrderedFields = difference(originalOrder, orderConfig.fields);
```

sandrina-p: Yes, let's first align the warn/error API in the other PR https://github.com/remoteoss/json-schema-form/pull/80#discussion_r1684195117

src/tests/modify.test.js
```diff
@@ -298,3 +306,87 @@ describe('modify() - basic mutations', () => {
...
+    };
+
+    const result = modify(baseExample, {
+      fields: {
+        address: {
+          order: (original) => {
+            return {
+              fields: original.reverse(), // ['city', 'zipcode', 'first_line']
+            };
+          },
```
danielcardoso5, line 373: It would be nice to include a test with the non-function version of order for fields' order

```
address: {
  order: {
    fields: ['city', 'zipcode', 'first_line']
  }
}
```

src/tests/modify.test.js
danielcardoso5: I prefer option A for its readability, it took me some time to wrap my head around option B and the nested `order` keys to change the root order look strange to me. 

> I think everything related to order, should be under order property

Definitely agree with this statement, and the `modify` code leaves anything related to `order` as the last mutations taking place.

Considering that, I think something along the lines @eng-almeida suggested in [this comment](https://github.com/remoteoss/json-schema-form/pull/78/files#r1677556530) could also work. Something like this:

```
const resultB = modify(baseExample, {
  order: {
     'root': (original) => {
        return original.reverse(); // ['age', 'address']
     },
     'address': (original) => {
         return  original.reverse(); // ['city', 'zipcode', 'first_line']
      },
     'some.nested.field': ['field1' ,'field2'], // function or array-order returns are possible for any key
    };
  },
});
```

danielcardoso5: I prefer option A for its readability, it took me some time to wrap my head around option B and the nested `order` keys to change the root order look strange to me. 

> I think everything related to order, should be under order property

Definitely agree with this statement, and the `modify` code leaves anything related to `order` as the last mutations taking place.

Considering that, I think something along the lines @eng-almeida suggested in [this comment](https://github.com/remoteoss/json-schema-form/pull/78/files#r1677556530) could also work. Something like this:

```
const resultB = modify(baseExample, {
  order: {
     'root': (original) => {
        return original.reverse(); // ['age', 'address']
     },
     'address': (original) => {
         return  original.reverse(); // ['city', 'zipcode', 'first_line']
      },
     'some.nested.field': ['field1' ,'field2'], // function or array-order returns are possible for any key
    };
  },
});
```

sandrina-p: @danielcardoso5 you say that you prefer approach A but then your comment follows approach B, I'm confused 😅

Also, in your code snippet `order.root` will clash with  JSON Schemas that have a field literally called `root` 🫠  Based on the other comments, the way to avoid it using `fields` as we do in `pick` and `create`, and use `root` as another reserved word:

```js
modify(schema, {
  order: {
    "root": ["name", "address", "birthdate"],
    fields: {
      'address': ['street', 'city', 'zip'],
      'some.nested.field': ['field1', 'field2', 'field3']
    }
  }
})
```

But again, this is approach B (all under `order` scope), not approach A (each fieldset with its own scope).

But... the reality is that even the other methods (eg to create, and to edit) follow an approach B. So, yeah, let's do it approach B here too.

Here's a practical example, where we'd like to change the `address` fieldset, to:
- edit address.zipcode to only allow 4 digits
- create a new field address.state
- move address.zipcode field to the first place:

```js
modify(schema, {
  fields: {
    'address.zip_code': {
      minLength: 4,
      maxLength: 4,
    }
  },
  create: {
    fields: {
      'address.state': {
        title: 'Your state'
      }
    }
  },
  order: {
    fields: {
      address: (original) => ['zip_code', ...original.filter(f => f !== 'zip_code')],
    },
  },
  onWarn: () => {...}
});
```

WDYT? 🤔

sandrina-p: I'm sorry but I'm not happy with the above, at all. The API is ok for root fields, but for fieldsets it's 💩 .

Then I remembered it's already possible to override the order of fieldsets with what we have in `main` LOL If you try this it works 😅 

```js
modify(schema, {
  fields: {
    'address': (attrs) => ({
      'x-jsf-order': ['zip_code', ...attrs['x-jsf-order'].filter(f => f !== 'zip_code')],
    })
  },
});
```

TLDR: `order()` is only meant root order. For nested use `config.fields` instead, until a real big usecase for a better API is needed. WDYT?

danielcardoso5: > TLDR: order() is only meant root order. For nested fields, use config.fields as above until a real big usecase for a better API is needed. WDYT?

Order is a property of a fieldset, so I think it is okay changing it via config.fields as you suggest.

The reason I prefer approach A and then suggested an approach following B is readability. Approach A was more readable, my suggestion was to make B more readable with less boilerplate since putting everything under `order` scope made more sense to me – especially because `order` mutations should be the last thing taking place. 

> it's already possible to modify the order of fieldsets with what we shipped to main

IMO this makes any approach above irrelevant since having two ways of accomplishing the same thing is no good.

I would make clear via the API that `order` is only for root order, you could name it `rootOrder`/`baseOrder`.

Based on the practical example you gave above, what do you think of this?

```
modify(schema, {
  fields: {
    "address.zip_code": {
      minLength: 4,
      maxLength: 4,
    },
    address: (attrs) => ({
      "x-jsf-order": ["zip_code", ...attrs["x-jsf-order"].filter((f) => f !== "zip_code")],
    }),
  },
  create: {
    fields: {
      "address.state": {
        title: "Your state",
      },
    },
  },
  rootOrder: ["address", "age"],
});
```

If we go with this approach we still need to be mindful of the order of operations. Mutating the `order` of fields should only take place after all the other mutations. This would require the initial `rewriteFields` to ignore any `x-jsf-order` attribute mutation, `reorderFields` would then consider that attribute with all the mutations that took place so far.

From an API interaction standpoint, the `x-jsf-order` attribute mutation might be a bit surprising when using the mutation fn pattern because it might not receive the original `x-jsf-order` value, e.g. fields were added/omitted to the fieldset by previous mutations. I don't think this is bad in any way though.

sandrina-p: > IMO this makes any approach above irrelevant since having two ways of accomplishing the same thing is no good.

This 100%. Less is more. Nested order removed then ^.^


> This would require the initial rewriteFields to ignore any x-jsf-order attribute mutation, reorderFields would then consider that attribute with all the mutations that took place so far.

Good point, I'll take that into account and add tests.

That's it then, I'll implement it now and get back to you when it's done.

src/tests/modify.test.js
```diff
@@ -397,3 +397,89 @@ describe('modify() - basic mutations', () => {
     });
   });
 });
+
+describe('modify() - reoder fields', () => {
```
sandrina-p, line 401: I've rebased this (Order PR) with main https://github.com/remoteoss/json-schema-form/pull/83, including the new specs discussed in the previous threads. Ready for a last review :)

src/modify.js
```diff
@@ -1,8 +1,10 @@
+import difference from 'lodash/difference';
 import get from 'lodash/get';
 import mergeWith from 'lodash/mergeWith';
 
 const WARNING_TYPES = {
   FIELD_TO_CHANGE_NOT_FOUND: 'FIELD_TO_CHANGE_NOT_FOUND',
+  ODER_MISSING_FIELDS: 'ODER_MISSING_FIELDS',
```
danielcardoso5, line 7: Typo

```suggestion
  ORDER_MISSING_FIELDS: 'ORDER_MISSING_FIELDS',
```

src/modify.js
```diff
@@ -92,20 +94,43 @@ function rewriteAllFields(schema, configCallback, context) {
   });
 }
 
+function reorderFields(schema, configOrder) {
+  if (!configOrder) return { warnings: null };
+
+  const warnings = [];
+  const originalOrder = schema['x-jsf-order'];
```
danielcardoso5, line 101: Is `x-jsf-order` a required property of a json-schema? If not we might need a fallback here.

sandrina-p: good point, added fallback and a test!

danielcardoso5: 🙇 🙇 🙇

PR #79, opened 2024-07-11
feat(modify): Allow picking fields

sandrina-p:
Follow-up of https://github.com/remoteoss/json-schema-form/pull/71, scoped on allowing to pick certain fields

[Linear issue (internal)](https://linear.app/remote/issue/PBYR-1371)

### New specs

Imagine the schema below with 4 fields to buy tickets:

```js
const schemaTickets = {
  properties: {
    age: {
      title: 'Age',
      type: 'integer',
    },
    quantity: {
      title: 'Quantity',
      type: 'integer',
    },
    has_premium: {
      title: 'Has premium',
      type: 'string',
    },
    premium_id: {
      title: 'Premium ID',
      type: 'boolean',
    },
  },
  'x-jsf-order': ['age', 'quantity', 'has_premium', 'premium_id'],
  allOf: [
    {
      if: {
        properties: {
          has_premium: {
            const: 'yes',
          },
        },
        required: ['has_premium'],
      },
      then: {
        required: ['premium_id'],
      },
      else: {
        properties: {
          premium_id: false,
        },
      },
    },
  ],
};
```

If you want a JSON Schema with just the `quantity` field, pass the following config:

```js
const { schema, warnings } = modify(schemaTickets, {
  pick: ['quantity'],
});
```

Which will return you the schema:

```js
{
  properties: {
    quantity: {
      title: 'Quantity',
      type: 'integer',
    }
  },
  'x-jsf-order': ['quantity'],
  allOf: undefined
};
```

#### Other features:

- When picked fields have conditional fields, they are also included and it's included in the `warnings`.
- NOT done:
    - Pick nested fields
    - Ignore conditionals with missing fields

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.10.1-beta.0",
+  "version": "0.11.0-dev.20240725074901",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.10.0-beta.0",
+  "version": "0.10.1-beta.0",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.0-dev.20240725074901",
+  "version": "0.10.0-beta.0",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/src/modify.js
+++ b/src/modify.js
@@ -92,7 +92,7 @@ function rewriteFields(schema, fieldsConfig) {
       mergeReplaceArray
     );
 
-    // recrusive
+    // recursive
     if (fieldChanges.properties) {
       const result = rewriteFields(get(schema.properties, fieldPath), fieldChanges.properties);
       warnings.push(result.warnings);
@@ -201,13 +201,11 @@ function pickFields(originalSchema, pickConfig) {
         newSchema[attrKey] = attrValue.filter((fieldName) => fieldsToPick.includes(fieldName));
         break;
       case 'allOf': {
-        const newAllOf = [];
-        // remove conditional ("if, then, else") if it does not contain any reference to fieldsToPick
-        originalSchema.allOf.forEach((condition) => {
-          if (isConditionalReferencingAnyPickedField(condition, fieldsToPick)) {
-            newAllOf.push(condition);
-          }
-        });
+        // remove conditionals that do not contain any reference to fieldsToPick
+        const newAllOf = originalSchema.allOf.filter((condition) =>
+          isConditionalReferencingAnyPickedField(condition, fieldsToPick)
+        );
+
         newSchema[attrKey] = newAllOf;
 
         break;
@@ -220,33 +218,36 @@ function pickFields(originalSchema, pickConfig) {
   // Look for unpicked fields in the conditionals...
   let missingFields = {};
 
-  newSchema.allOf.forEach((condition, ix) => {
+  newSchema.allOf.forEach((condition) => {
     const { if: ifCondition, then: thenCondition, else: elseCondition } = condition;
-
+    const index = originalSchema.allOf.indexOf(condition);
     missingFields = {
       ...missingFields,
       ...findMissingFields(ifCondition, {
         fields: fieldsToPick,
-        path: `allOf[${ix}].if`,
+        path: `allOf[${index}].if`,
       }),
       ...findMissingFields(thenCondition, {
         fields: fieldsToPick,
-        path: `allOf[${ix}].then`,
+        path: `allOf[${index}].then`,
       }),
       ...findMissingFields(elseCondition, {
         fields: fieldsToPick,
-        path: `allOf[${ix}].else`,
+        path: `allOf[${index}].else`,
       }),
     };
   });
 
   if (Object.keys(missingFields).length > 0) {
-    // Read them to the new schema...
+    // Re-add them to the schema...
     Object.entries(missingFields).forEach(([fieldName]) => {
       set(newSchema.properties, fieldName, originalSchema.properties[fieldName]);
     });
     // And warn about it (the most important part!)
-    pickConfig.onWarn(missingFields);
+    pickConfig.onWarn({
+      message: 'You picked a field which has related conditional fields. They got added:',
+      missingFields,
+    });
   }
 
   return newSchema;
@@ -267,18 +268,12 @@ function findMissingFields(conditional, { fields, path }) {
     }
   });
 
-  Object.entries(conditional.properties || []).forEach(([fieldName, fieldAttrs]) => {
+  Object.entries(conditional.properties || []).forEach(([fieldName]) => {
     if (!fields.includes(fieldName)) {
       missingFields[fieldName] = { path };
     }
 
-    if (fieldAttrs.properties) {
-      const nested = findMissingFields(fieldAttrs.properties, fields);
-      missingFields = {
-        ...missingFields,
-        ...nested,
-      };
-    }
+    // TODO support nested fields (eg if properties.adddress.properties.door_number)
   });
 
   return missingFields;
--- a/src/modify.js
+++ b/src/modify.js
@@ -1,6 +1,7 @@
+import difference from 'lodash/difference';
 import get from 'lodash/get';
-import merge from 'lodash/merge';
 import intersection from 'lodash/intersection';
+import merge from 'lodash/merge';
 import mergeWith from 'lodash/mergeWith';
 import set from 'lodash/set';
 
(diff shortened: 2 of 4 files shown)


src/modify.js
```diff
@@ -12,14 +14,46 @@ function shortToFullPath(path) {
 }
 
 function standardizeAttrs(attrs) {
-  const { errorMessage, properties, ...rest } = attrs;
+  const {
+    errorMessage, // to be renamed
+    properties, // ignored because of recurisve call
```
danielcardoso5, line 19:
```suggestion
    properties, // ignored because of recursive call
```

src/modify.js
```diff
@@ -66,12 +101,122 @@ function rewriteAllFields(schema, configCallback, context) {
...
+        const newAllOf = [];
+        // remove conditional ("if, then, else") if it does not contain any reference to fieldsToPick
+        originalSchema.allOf.forEach((condition) => {
+          if (isConditionalReferencingAnyPickedField(condition, fieldsToPick)) {
+            newAllOf.push(condition);
+          }
+        });
+        newSchema[attrKey] = newAllOf;
+
+        break;
```
danielcardoso5, line 134: **nit** we can avoid the reassignment

```suggestion
        // remove conditional ("if, then, else") if it does not contain any reference to fieldsToPick
        const newAllOf = originalSchema.allOf.filter((condition) =>
          isConditionalReferencingAnyPickedField(condition, fieldsToPick)
        );
        
        newSchema[attrKey] = newAllOf;
        break;
```

src/modify.js
```diff
@@ -66,12 +101,122 @@ function rewriteAllFields(schema, configCallback, context) {
...
+      }),
+      ...findMissingFields(elseCondition, {
+        fields: fieldsToPick,
+        path: `allOf[${ix}].else`,
+      }),
+    };
+  });
+
+  if (Object.keys(missingFields).length > 0) {
+    // Read them to the new schema...
```
danielcardoso5, line 165: I think this means `Add`

```suggestion
    // Add them to the new schema...
```

src/modify.js
```diff
@@ -66,12 +101,122 @@ function rewriteAllFields(schema, configCallback, context) {
...
+    }
+  });
+
+  Object.entries(conditional.properties || []).forEach(([fieldName, fieldAttrs]) => {
+    if (!fields.includes(fieldName)) {
+      missingFields[fieldName] = { path };
+    }
+
+    if (fieldAttrs.properties) {
+      const nested = findMissingFields(fieldAttrs.properties, fields);
```
danielcardoso5, line 200: Is this correct? Doesn't `findMissingFields` expect a conditional? I'm a bit lost by what `fieldAttrs.properties` refers to in this context since we're already going through `conditional.properties` above.

src/modify.js
```diff
@@ -66,12 +101,122 @@ function rewriteAllFields(schema, configCallback, context) {
...
+  });
+
+  if (Object.keys(missingFields).length > 0) {
+    // Read them to the new schema...
+    Object.entries(missingFields).forEach(([fieldName]) => {
+      set(newSchema.properties, fieldName, originalSchema.properties[fieldName]);
+    });
+    // And warn about it (the most important part!)
+    pickConfig.onWarn({
+      message: 'You picked a field with conditional fields. They got added:',
```
danielcardoso5, line 171: **suggestion**

```suggestion
      message: 'You picked a field which has related conditional fields. They got added:',
```

src/modify.js
```diff
@@ -66,12 +101,122 @@ function rewriteAllFields(schema, configCallback, context) {
...
+  let missingFields = {};
+
+  newSchema.allOf.forEach((condition, ix) => {
+    const { if: ifCondition, then: thenCondition, else: elseCondition } = condition;
+
+    missingFields = {
+      ...missingFields,
+      ...findMissingFields(ifCondition, {
+        fields: fieldsToPick,
+        path: `allOf[${ix}].if`,
```
danielcardoso5, line 151: This `ix` refers to the index of the `newSchema.allOf` and not the `allOf` from the original schema. I think we want to refer to the original index when logging, if that's the case you can use `originalSchema.allOf.indexOf(condition)`.

src/modify.js
```diff
@@ -66,12 +101,122 @@ function rewriteAllFields(schema, configCallback, context) {
...
+  });
+
+  // Look for unpicked fields in the conditionals...
+  let missingFields = {};
+
+  newSchema.allOf.forEach((condition, ix) => {
+    const { if: ifCondition, then: thenCondition, else: elseCondition } = condition;
+
+    missingFields = {
+      ...missingFields,
```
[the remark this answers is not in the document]
sandrina-p: Oh yes we can! I simplified the code so much but forgot that part, thanks!

src/modify.js
sandrina-p: Oh god, it's official. I don't know how to type recurisive.

src/modify.js
sandrina-p: Let me get back to you first thing tomorrow morning to clarify this and apply all the other changes

src/modify.js
sandrina-p: Really nice catch! I modified the test case to catch this one.

src/modify.js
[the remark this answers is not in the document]
sandrina-p: Hmm, I looked again, actually we need to keep this one because it's a loop, otherwise it will ignore any findings from previous conditionals. I added a test to show you, if you delete this line you'll see it failing

src/modify.js
sandrina-p: Oh this one is to be recursive (\*doub-checked if word is well written\*), and nested fields, for example:

```js
{
  if: {
    properties: {
      address: {
        properties: {
          number: {
            const: 5
          }
        }
      }
    }
  },
  then: { /*...*/ },
  else: { /* ... */ }
}
```

I'll add a test in 1h after my call now.

sandrina-p: Actually, supporting conditionals with nested fields (fieldsets) has its own edge-cases too: It also needs to take into account the nested `allOf` (eg `properties.address.allOf`).

As I don't know of any case needing that right now, and I'm short on time, I rather leave it out for another day.

 I added more details to the current TODOs in the tests files.

In other words, I completely removed this block of code.

sandrina-p:
@danielcardoso5  Ready for a new review :D

src/modify.js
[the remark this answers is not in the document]
danielcardoso5: ah, apparently my brain circuitry forgot to deal to re-assignments 😅 we could use `Array.reduce` here then, but no need to change.

src/modify.js
danielcardoso5: Sounds good, thank you!

danielcardoso5 (approved):
Looks good @sandrina-p! Thank you for the added comments re: fieldsets ✨

src/tests/modify.test.js
```diff
@@ -586,3 +649,118 @@ describe('modify() - create fields', () => {
     ]);
   });
 });
+
+describe('modify() - pick fields', () => {
+  it('basic usage', () => {
+    const { schema, warnings } = modify(schemaTickets, {
```
sandrina-p, line 655: I've rebased this (Pick PR) with main https://github.com/remoteoss/json-schema-form/pull/83, including the new specs discussed in the previous threads. Ready for a last review :)

src/modify.js
```diff
@@ -150,6 +180,110 @@ function createFields(schema, fieldsConfig) {
...
+    // Re-add them to the schema...
+    Object.entries(missingFields).forEach(([fieldName]) => {
+      set(newSchema.properties, fieldName, originalSchema.properties[fieldName]);
+    });
+
+    warnings.push({
+      type: WARNING_TYPES.PICK_MISSED_FIELD,
+      message: `The picked fields have related conditional fields that got added automatically. ${Object.keys(
+        missingFields
+      ).join(', ')}. Check "meta" for more details.`,
```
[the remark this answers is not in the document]
sandrina-p: done :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.10.1-beta.0",
+  "version": "0.10.0-beta.0",
```
danielcardoso5, line 3: Why the version revert here? Isn't `0.10.1-beta.0` published? Complete newbie here 😅

danielcardoso5 (approved):

package.json
sandrina-p: I was looking into the same, this is a bug (!) with the script to revert a dev version, and with the lint to check if the current version is okay (if you notice the pipeline passes when it should fail https://github.com/remoteoss/json-schema-form/actions/runs/10096419678/job/27919143600?pr=79)

The ideal flow:
1. The PR is updated with the main (have the latest version).
2. You release a dev version
3. The CI lint fails because the version is not a dev one.
4. You revert the dev version (it's a script that reverts the file change)
5. The CI lint passes because the version is restored.
6. You merge the PR and release a new version (beta) in main.


This flow:
1. The PR is outdated with main (it is still 10.0 instead of 10.1)
2. You release a dev version 
3. The CI lint fails because the version is not a dev one.
4. You revert the dev version (it's a script that reverts the file change)
5. The CI lint passes because the version is restored. (bug! 🐛 it should fail instead because the version was changed. Or at least fail because the MR is not updated with main) 
6. You merge the PR and release a new version (beta) in main.

Is this clear? 😅 Anyway, I manually fixed the version (rebased with main) and will open an issue to fix the CI lint scripts.

sandrina-p: Actually... this branch was already updated with main (10.1.0), so it's a bug with the step 4 (revert the dev version!) (commit https://github.com/remoteoss/json-schema-form/pull/79/commits/62b6b27982fb2b34371642d65c642aba875acfcd). Anyway, a problem for another day.

sandrina-p: In the meantime, https://github.com/remoteoss/json-schema-form/pull/81 just got merged, it has conflicts, let me fix them in a minute.

danielcardoso5 (dismissed):

package.json
sandrina-p: Yes, probably. I added this thread as reference for when someone picks it up :)

danielcardoso5 (approved):

PR #80, opened 2024-07-18
feat(modify): Allow creating fields

sandrina-p:
Follow-up of https://github.com/remoteoss/json-schema-form/pull/71, scoped on allowing to add new fields

[Linear issue (internal)](https://linear.app/remote/issue/EXPM-631)

### New specs

Imagine the schema with 2 fields:

```js
const baseSchema = {
  properties: {
    age: {
      title: 'Age',
      type: 'integer',
    },
    family_name: {
      title: 'Family Name',
      type: 'string',
    }
  },
  'x-jsf-order': ['age', 'family_name'],
};
```

If you want to add a new field to the JSON Schema pass the following config:

```js
const { schema }  = modify(baseSchema, {
  create: {
    username: {
      title: 'Username',
      type: 'string',
    },
  }
)
```

Which will return the schema:

```js
{
  properties: {
    age: {
      title: 'Age',
      type: 'integer',
    },
    family_name: {
      title: 'Family Name',
      type: 'string',
    },
    username: {
      title: 'Username',
      type: 'string',
    }
  },
  'x-jsf-order': ['age', 'family_name'],
};
```

### Creating an existing field
If you try to create a field that already exists, it get's ignored and it returns a `warning`. 

Here's an example with the `age` field based on the `baseSchema` above. 

```js
const { schema, warnings } = modify(baseSchema, {
   create: {
     age: { title: "Your age" }
  }
})

console.log(warnings):

[{
  type: "FIELD_TO_CREATE_EXISTS",
  message: `Creating field "age" was ignored because it already exists.`,
}];
```

Check tests for more examples.

### Other features:
- Support adding nested fields
- It supports reordering new fields (check tests and MR https://github.com/remoteoss/json-schema-form/pull/78)

### DIFF
--- a/src/modify.js
+++ b/src/modify.js
@@ -132,8 +132,8 @@ function createFields(schema, fieldsConfig) {
 
     const fieldInSchema = get(schema.properties, fieldPath);
     if (fieldInSchema) {
-      // NOTE: Not sure if we should ignore or overwrite it
-      // when it exists. Let's ignore until someone complains.
+      // Do not override/edit a field that already exists.
+      // That's the job of config.fields method.
       return;
     }
 
--- a/src/modify.js
+++ b/src/modify.js
@@ -7,6 +7,7 @@ import set from 'lodash/set';
 const WARNING_TYPES = {
   FIELD_TO_CHANGE_NOT_FOUND: 'FIELD_TO_CHANGE_NOT_FOUND',
   ORDER_MISSING_FIELDS: 'ORDER_MISSING_FIELDS',
+  FIELD_TO_CREATE_EXISTS: 'FIELD_TO_CREATE_EXISTS',
 };
 /**
  *
@@ -118,28 +119,36 @@ function reorderFields(schema, configOrder) {
 }
 
 function createFields(schema, fieldsConfig) {
-  if (!fieldsConfig) return null;
+  if (!fieldsConfig) return { warnings: null };
 
-  const fieldsToAdd = Object.entries(fieldsConfig);
+  const warnings = [];
+  const fieldsToCreate = Object.entries(fieldsConfig);
 
-  fieldsToAdd.forEach(([shortPath, fieldAttrs]) => {
+  fieldsToCreate.forEach(([shortPath, fieldAttrs]) => {
     const fieldPath = shortToFullPath(shortPath);
 
     if (fieldAttrs.properties) {
       // Recursive to nested fields...
-      createFields(get(schema.properties, fieldPath), fieldAttrs.properties);
+      const result = createFields(get(schema.properties, fieldPath), fieldAttrs.properties);
+      warnings.push(result.warnings);
     }
 
     const fieldInSchema = get(schema.properties, fieldPath);
-    if (fieldInSchema) {
-      // Do not override/edit a field that already exists.
-      // That's the job of config.fields method.
+    const { properties, ...otherAttrs } = fieldAttrs;
+
+    if (fieldInSchema && otherAttrs) {
+      warnings.push({
+        type: WARNING_TYPES.FIELD_TO_CREATE_EXISTS,
+        message: `Creating field "${shortPath}" was ignored because it already exists.`,
+      });
       return;
     }
 
     const fieldInObjectPath = set({}, fieldPath, standardizeAttrs(fieldAttrs));
     merge(schema.properties, fieldInObjectPath);
   });
+
+  return { warnings: warnings.flat() };
 }
 
 export function modify(originalSchema, config) {
@@ -149,7 +158,7 @@ export function modify(originalSchema, config) {
   const resultRewrite = rewriteFields(schema, config.fields);
   rewriteAllFields(schema, config.allFields);
 
-  createFields(schema, config.create);
+  const resultCreate = createFields(schema, config.create);
 
   const resultReorder = reorderFields(schema, config.orderRoot);
 
@@ -159,7 +168,9 @@ export function modify(originalSchema, config) {
     );
   }
 
-  const warnings = [resultRewrite.warnings, resultReorder.warnings].flat().filter(Boolean);
+  const warnings = [resultRewrite.warnings, resultCreate.warnings, resultReorder.warnings]
+    .flat()
+    .filter(Boolean);
 
   return {
     schema,
--- a/src/modify.js
+++ b/src/modify.js
@@ -117,7 +117,7 @@ function reorderFields(schema, configOrder) {
   return { warnings };
 }
 
-function setFields(schema, fieldsConfig) {
+function createFields(schema, fieldsConfig) {
   if (!fieldsConfig) return null;
 
   const fieldsToAdd = Object.entries(fieldsConfig);
@@ -127,7 +127,7 @@ function setFields(schema, fieldsConfig) {
 
     if (fieldAttrs.properties) {
       // Recursive to nested fields...
-      setFields(get(schema.properties, fieldPath), fieldAttrs.properties);
+      createFields(get(schema.properties, fieldPath), fieldAttrs.properties);
     }
 
     const fieldInSchema = get(schema.properties, fieldPath);
@@ -148,7 +148,8 @@ export function modify(originalSchema, config) {
 
   const resultRewrite = rewriteFields(schema, config.fields);
   rewriteAllFields(schema, config.allFields);
-  setFields(schema, config.add);
+
+  createFields(schema, config.create);
 
   const resultReorder = reorderFields(schema, config.orderRoot);
 
--- a/src/modify.js
+++ b/src/modify.js
@@ -23,12 +23,11 @@ function mergeReplaceArray(_, newVal) {
 }
 
 function standardizeAttrs(attrs) {
-  const { errorMessage, presentation, properties, ...rest } = attrs;
+  const { errorMessage, properties, ...rest } = attrs;
 
   return {
     ...rest,
     ...(errorMessage ? { 'x-jsf-errorMessage': errorMessage } : {}),
-    ...(presentation ? { 'x-jsf-presentation': presentation } : {}),
   };
 }
 
@@ -125,8 +124,6 @@ function setFields(schema, fieldsConfig) {
 
   fieldsToAdd.forEach(([shortPath, fieldAttrs]) => {
     const fieldPath = shortToFullPath(shortPath);
-
-    console.log('fieldAttrs', fieldAttrs);
 
     if (fieldAttrs.properties) {
       // Recursive to nested fields...
--- a/src/modify.js
+++ b/src/modify.js
@@ -134,9 +134,8 @@ function createFields(schema, fieldsConfig) {
     }
 
     const fieldInSchema = get(schema.properties, fieldPath);
-    const { properties, ...otherAttrs } = fieldAttrs;
 
-    if (fieldInSchema && otherAttrs) {
+    if (fieldInSchema) {
       warnings.push({
         type: WARNING_TYPES.FIELD_TO_CREATE_EXISTS,
         message: `Creating field "${shortPath}" was ignored because it already exists.`,
--- a/src/modify.js
+++ b/src/modify.js
@@ -1,6 +1,8 @@
 import difference from 'lodash/difference';
 import get from 'lodash/get';
+import merge from 'lodash/merge';
 import mergeWith from 'lodash/mergeWith';
+import set from 'lodash/set';
 
 const WARNING_TYPES = {
   FIELD_TO_CHANGE_NOT_FOUND: 'FIELD_TO_CHANGE_NOT_FOUND',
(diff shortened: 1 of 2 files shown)


src/tests/modify.test.js
```diff
@@ -298,3 +313,90 @@ describe('modify() - basic mutations', () => {
     });
   });
 });
+
+describe('modify() - set new fields', () => {
+  it('add base field', () => {
+    const result = modify(schemaAddress, {
+      add: {
```
sandrina-p, line 320: note: Regarding the "add" name I also thought about:
- `addField`, but we already have the other shorter names like `pick`, `omit` and `order`. So `add` seems better.
- `set`, but it's not "set" really, because "set" implies replacing existing fields, which does not do.
- `insert`, does not make sense because it implies a specific position, which in this case is not guaranteed (that's up to `order`)

What are your thoughts?

src/tests/modify.test.js
```diff
@@ -257,6 +257,21 @@ describe('modify() - basic mutations', () => {
     });
   });
 
+  it('unexistent fields gets ignored', () => {
```
sandrina-p, line 260: Note on why I decided to write this explicit test:

In Remote we have a lot of global customization that run equally across multiple different JSON Schemas.
But we only want to apply the customizations relative to existing fields. Otherwise we'd inject (create) new fields even though they don't exist. That's why we have the new method on this PR, specific to create new fields.

src/modify.js
```diff
@@ -66,12 +67,39 @@ function rewriteAllFields(schema, configCallback, context) {
...
+    const fieldPath = shortToFullPath(shortPath);
+
+    if (fieldAttrs.properties) {
+      // Recursive to nested fields...
+      setFields(get(schema.properties, fieldPath), fieldAttrs.properties);
+    }
+
+    const fieldInSchema = get(schema.properties, fieldPath);
+    if (fieldInSchema) {
+      // NOTE: Not sure if we should ignore or overwrite it
```
sandrina-p, line 85: Note on why I decided to ignore modifying existing fields: Read this comment first https://github.com/remoteoss/json-schema-form/pull/80/files#r1683547346.

So if `fields` is exclusive for existing fields, then `create` is exclusive for new fields. Does it make sense?

src/tests/modify.test.js
sandrina-p: Update: Eureka moment while writing the other comments, I'll rename this to `create` :D

src/tests/modify.test.js
```diff
@@ -298,3 +317,90 @@ describe('modify() - basic mutations', () => {
     });
   });
 });
+
+describe('modify() - set new fields', () => {
```
danielcardoso5, line 321:
```suggestion
describe('modify() - create new fields', () => {
```

src/modify.js
```diff
@@ -66,12 +67,39 @@ function rewriteAllFields(schema, configCallback, context) {
...
+
+    if (fieldAttrs.properties) {
+      // Recursive to nested fields...
+      createFields(get(schema.properties, fieldPath), fieldAttrs.properties);
+    }
+
+    const fieldInSchema = get(schema.properties, fieldPath);
+    if (fieldInSchema) {
+      // Do not override/edit a field that already exists.
+      // That's the job of config.fields method.
```
danielcardoso5, line 86: Similar to the pick fields, should we warn when a field creation is not done?

Something like:
```
  pickConfig.onWarn({
    message: `You've tried to create a field at ${fieldPath}, however the field path already exists in the schema. This field will be ignored.'
  });
```

src/tests/modify.test.js
danielcardoso5: ahah `create` sounds good! Also thought of `inject` since you're not only creating the field property but "injecting" the field and it's properties into an existing schema

sandrina-p: we are not doing rocket science, no need to inject 🚀 ahahah

src/modify.js
sandrina-p: Ah good point! But now the question is the API... Where exactly do we put `onWarn`?

In the pick, the API is:

```js
modify(schema, {
  pick: {
    fields: {
        [fieldsName]: "...."
    },
    onWarn: () => "...."
  }
})
```

But the API for create was different. It accepts the fields directly. Where would we put the `onWarn()`?

```js
modify(schema, {
  create: {
    [fieldsName]: "...."
  }
})
```

Or, we could have a global onWarn to all type of warnings. 

```js
modify(schema, {
  create: {
    [fieldsName]: "...."
  },
  pick: { ... }
  onWarn: (warnings) =>
})
```

The warnings would be an array  of all multiple warnings, from pick, create, etc etc. WDYT?

sandrina-p: The next question is. Shall we go back to the pick MR https://github.com/remoteoss/json-schema-form/pull/79/files and change the API to be flat like `create` and `fields`?

Before:

```js
modify(schema, {
  pick: {
    fields: {
        [fieldsName]: "...."
    },
    onWarn: () => "...."
  }
})
```

After:

```js
modify(schema, {
  pick: {
    [fieldsName]: "...."
  },
  onWarn: () => "...."
  }
})
```

sandrina-p: Actually, the order PR https://github.com/remoteoss/json-schema-form/pull/78 also has nested `fields`:

```
const result = modify(schema, {
  order: {
    fields: ['field_c', 'field_a', 'field_b'],
    remaining: 'end'
  },
);
```

Decisions, decisions... I'm towards adding the nested `fields` attr to every main method. (create, pick, order).

@joaoalmeida-remote

src/tests/modify.test.js
eng-almeida: @sandrina-p just letting you know that the PR description is not updated with `create`

sandrina-p: @eng-almeida updated :)

src/modify.js
eng-almeida: I agree having `fields` property across all the utility methos.
As for the on warn we could follow the usal Node JS function signature and accept a onError parameter (as a 3rd parameter) and so it would look like:

```
modify(schema, {
  pick: {
    fields: {
        [fieldsName]: "...."
    },
  },
  (err) => {
    console.log(err)
  } 
})
```

One of the advantages is that we could use the onError method in all the utilities methods. Thoughts?

danielcardoso5: @eng-almeida you meant as a third parameter to the `modify` call right? So like this?

```
modify(schema, {
  pick: {
    fields: {
        [fieldsName]: "...."
    },
  }
}, (err) => { console.log(err)} )
```

I'm not sure here, putting it inside the config object gives more flexibility to later – if necessary – change the `modify` API or introduce a new callback. You can make changes to the config without consumers having to modify calls if they don't want the new behavior, e.g. deprecated a config option.

Also not sure about `onError` since we don't seem to error ever, we just warn about potentially mis-configured schemas.

sandrina-p: I agree with @danielcardoso5, in other words:

```js
modify(schema, {
  pick: {
    fields: {
        [fieldsName]: "...."
    },
  },
  onWarn: => {}
}
```

We don't deal with errors at all at the moment, but if we ever did, probably it's gonna be a typical try catch. 

Or, another point different from `createHeadlessForm`, there we do like this:

```js
const { fields, error } = createHeadlessForm(schema)
```

Perhaps we should do the same for modify?

```js
const { schema, error } = modify(schema, originalSchema)
```

sandrina-p: Do you know if there are major pros/cons of using a callback vs direct result, besides the way you access/handle it?

As I see it, with callback we can enforce it's not forgotten (TS, throwing error, etc). With the direct result, we can't.

I'm not sure which one to pick, need to give it a thought. But probably would keep it with the callback.(the first snippet in my previous comment)

eng-almeida: > you meant as a third parameter to the modify call right

@danielcardoso5 yes, thanks for the correction 🙌  I agree with the flexibility argument, even though IMO it's ok to release major versions with breaking changes. My only concern with the initial proposal is with the API format by adding `onWarn` callback next to the fields. IMO It doesn't feel very intuitive.

> I'm not sure which one to pick, need to give it a thought. 

If we're looking for consistency (with createHeadlessForm) maybe `const { schema, error, warn } = modify(schema, originalSchema)` is the best option.

sandrina-p: > even though IMO it's ok to release major versions with breaking changes.

Yeah, I'm okay with breaking changes too, this is still in beta.

> If we're looking for consistency (with createHeadlessForm)

I don't see it as a mandatory thing, to be honest. I prioritize quality over consistency among these 2 modules

sandrina-p: Let me call @brennj here too. If you need a recap, let us know!

danielcardoso5: > maybe const { schema, error, warn } = modify(schema, originalSchema) is the best option.

Agree! Another reason I favor this approach over the callback pattern is that you only call `modify` once and you know the warnings/errors as soon as you make this call. The `onWarn` kind of callback pattern suggests you could potentially receive more warnings after the `modify` call and you could "listen" to them, but that's not the case.

danielcardoso5: > As I see it, with callback we can enforce it's not forgotten (TS, throwing error, etc). With the direct result, we can't.

I think this is the main drawback of using the return approach. We can always issue a warning whenever someone calls `modify` (and give the ability to suppress it) to inform that calls to `modify` can result in invalid schemas and that we provide a `warnings` array that they can use to be informed of things that were potentially overlooked. 

Since we don't have TS right now we would need to either throw an error or warn that the `onWarn` callback was not provided.

sandrina-p: > The `onWarn` callback suggests you could potentially receive more warnings

100% true, this is not an async operation, so one less reason to use it!

> Since we don't have TS right now 

We do have it, kind of... we have the Type declarations file. Either way, let's go then with the returned value approach then. (this will be a breaking change). I'll make sure to document `warn` very well in the docs.

> We can always issue a warning whenever someone calls modify (and give the ability to suppress it) 

What do you mean here with "suppress"?

danielcardoso5: > What do you mean here with "suppress"?

So what I was suggesting was to log a warning whenever there's a call to `modify`, something along the lines of:

`When using `modify` make sure you log the returned `warnings` as they provide useful debugging information…`

If the developers are already using the `warnings` array, there's no point in keeping logging that warning all the time, so they could potentially disable it by passing a flag to `modify`, example:

```
modify(schema, {
  pick: ...,
  disableLogging: true 
})
```

But now this made me realize that we can just log any warnings internally within `modify` by default, without having devs to either use a callback or the `warnings` return – although we can still provide the latter. If they really want to disable the built-in logs in local development, they could pass `disableLogging: true`. Most bundling tools already strip any logging in production builds.

sandrina-p: Oh I see... well, I think that will be super noisy but for a good cause.💡 I'll create an issue to tackle later, and focus now on finishing the current PRs.

I'll let you know when it's ready for a final review!

brennj: Just chiming in from the tag, trying to get proper understanding of the thread, I believe I like what you're all proposing with returning `warn` etc from `modify`.

src/tests/modify.test.js
```diff
@@ -397,3 +397,118 @@ describe('modify() - basic mutations', () => {
     });
   });
 });
+
+describe('modify() - create new fields', () => {
+  it('create base field', () => {
+    const result = modify(schemaAddress, {
+      create: {
```
sandrina-p, line 404: I've rebased this (Create PR) with main https://github.com/remoteoss/json-schema-form/pull/83, including the new specs discussed in the previous threads. Ready for a last review :)

sandrina-p: (FYI: I have a call now, will be back in 1h at ~15 UTC.) Then I'll finish updating the last PR #79

src/modify.js
```diff
@@ -92,20 +95,55 @@ function rewriteAllFields(schema, configCallback, context) {
...
+    if (fieldAttrs.properties) {
+      // Recursive to nested fields...
+      const result = createFields(get(schema.properties, fieldPath), fieldAttrs.properties);
+      warnings.push(result.warnings);
+    }
+
+    const fieldInSchema = get(schema.properties, fieldPath);
+    const { properties, ...otherAttrs } = fieldAttrs;
+
+    if (fieldInSchema && otherAttrs) {
```
danielcardoso5, line 116: Why do we need to check for `otherAttrs` here?

src/tests/modify.test.js
danielcardoso5: Changes look good, just have a question but not a blocker! (I'll be out between ~15-17UTC, leaving this approved!)

danielcardoso5 (approved):

src/modify.js
sandrina-p: This one is tough to explain. Basically, we have 2 types of changes to fieldsets:
- the fieldset itself (eg the "address" title — those are is "otherAttrs")
- a field inside the fieldset. (eg address.street — those are the "properties").

And the initial implemented needed this, so we could distinguish when to ignore it or not. 

Then I refactored the order of operations in this function to not be needed but forgot to add this additional check. Removed!

danielcardoso5 (approved):

src/tests/modify.test.js
sandrina-p: Thanks! I removed the last piece that was not needed. @joaoalmeida-remote could you re-approve?

sandrina-p: With the #78 merged, I rebased this PR with main to solve conflicts, and deleted the last test (it was skipped) as I realized we didn't need it anyway. [a33dd02](https://github.com/remoteoss/json-schema-form/pull/80/commits/a33dd02dbdc79ea0e8ee87698fcfa11a1bc71b48)

eng-almeida (approved):

PR #83, opened 2024-07-24
feat(modify): BREAKING CHANGE - Change the return result to include warnings

sandrina-p:
Follow-up of https://github.com/remoteoss/json-schema-form/pull/71.

[Linear issue (internal)](https://linear.app/remote/issue/pbyr-1417)

Now the result is an object with both `schema` and `warn`. The `warn` will be an array of all warnings highlighting unexpected changes. This can help you spot and avoid possible bugs with your modifications.

### Before

```js
const schema = modify(yourSchema, config)
```

### Now

```js
const { schema, warn } = modify(yourSchema, config)
```

When using `modify()`, it will also log a `console.log` by default calling you out to handle the `warn` on your behalf. For example, report the `warn` to your error tracking systems.

You can disable the log by passing `muteWarningTip`

```js
const { schema, warn } = modify(yourSchema, {
  muteWarningTip: true,
})
```

Full discussion in another PR https://github.com/remoteoss/json-schema-form/pull/80#discussion_r1684195117

### DIFF
--- a/src/modify.js
+++ b/src/modify.js
@@ -80,10 +80,22 @@ function rewriteAllFields(schema, configCallback, context) {
 
 export function modify(originalSchema, config) {
   const schema = JSON.parse(JSON.stringify(originalSchema));
+  let warn = []; // To be implemented in next PRs.
 
+  // All these functions mutate "schema",
+  // that's why we create a copy above
   rewriteFields(schema, config.fields);
 
   rewriteAllFields(schema, config.allFields);
 
-  return schema;
+  if (!config.muteWarningTip) {
+    console.warn(
+      'json-schema-form modify(): Make sure you log the returned `warn` as they highlight possible bugs in your modifications. To mute this log, pass `muteWarningTip: true` to the config.'
+    );
+  }
+
+  return {
+    schema,
+    warn: warn.length > 0 ? warn : undefined,
+  };
 }
--- a/src/tests/modify.test.js
+++ b/src/tests/modify.test.js
@@ -102,6 +102,36 @@ const schemaAddress = {
   },
 };
 
+beforeAll(() => {
+  jest.spyOn(console, 'warn').mockImplementation(() => {});
+});
+
+afterAll(() => {
+  console.warn.mockRestore();
+});
+
+describe('modify() - warnings', () => {
+  it('logs a warning by default', () => {
+    const result = modify(schemaPet, {});
+
+    expect(console.warn).toBeCalledWith(
+      'json-schema-form modify(): Make sure you log the returned `warn` as they highlight possible bugs in your modifications. To mute this log, pass `muteWarningTip: true` to the config.'
+    );
+
+    console.warn.mockClear();
+    expect(result.warn).toBeUndefined();
+  });
+
+  it('given muteWarningTip, it does not log the warning', () => {
+    const result = modify(schemaPet, {
+      muteWarningTip: true,
+    });
+
+    expect(console.warn).not.toBeCalled();
+    expect(result.warn).toBeUndefined();
+  });
+});
+
 describe('modify() - basic mutations', () => {
   it('replace base field', () => {
     const result = modify(schemaPet, {
@@ -119,7 +149,7 @@ describe('modify() - basic mutations', () => {
       },
     });
 
-    expect(result).toMatchObject({
+    expect(result.schema).toMatchObject({
       properties: {
         pet_name: {
           title: 'Your pet name',
@@ -135,24 +165,30 @@ describe('modify() - basic mutations', () => {
   it('replace nested field', () => {
     const result = modify(schemaAddress, {
       fields: {
+        // You can use dot notation
         'address.street': {
           title: 'Street name',
         },
         'address.city': () => ({
           title: 'City name',
         }),
-        address: () => {
+        // Or pass the native object
+        address: (fieldAttrs) => {
           return {
             properties: {
-              number: { title: 'Door number' },
+              number: (nestedAttrs) => {
+                return {
+                  'x-test-siblings': Object.keys(fieldAttrs.properties),
+                  title: `Door ${nestedAttrs.title}`,
+                };
+              },
             },
           };
         },
-        // TODO: write test to nested field
       },
     });
 
-    expect(result).toMatchObject({
+    expect(result.schema).toMatchObject({
       properties: {
         address: {
           properties: {
@@ -160,7 +196,8 @@ describe('modify() - basic mutations', () => {
               title: 'Street name',
             },
             number: {
-              title: 'Door number',
+              title: 'Door Number',
+              'x-test-siblings': ['street', 'number', 'city'],
             },
             city: {
               title: 'City name',
@@ -188,7 +225,7 @@ describe('modify() - basic mutations', () => {
       },
     });
 
-    expect(result).toMatchObject({
+    expect(result.schema).toMatchObject({
       properties: {
         has_pet: {
           dataFoo: 'abc',
@@ -203,6 +240,7 @@ describe('modify() - basic mutations', () => {
           styleDecimals: 2,
         },
         pet_address: {
+          // assert recursivity
           properties: {
             street: {
               dataFoo: 'abc',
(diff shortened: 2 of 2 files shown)


src/modify.js
```diff
@@ -80,10 +80,22 @@ function rewriteAllFields(schema, configCallback, context) {
...
   const schema = JSON.parse(JSON.stringify(originalSchema));
+  let warn = []; // To be implemented in next PRs.
 
+  // All these functions mutate "schema",
+  // that's why we create a copy above
   rewriteFields(schema, config.fields);
 
   rewriteAllFields(schema, config.allFields);
 
-  return schema;
```
sandrina-p, line 88: I decided to create a specific PR for this breaking change. Once merged, all the the other open PRs (order, pick, create) are easier to work on top of.

The new version (10.1.0) will only be published when all the other 3 PRs are approved.

Does that make sense to you?

src/modify.js
```diff
@@ -80,10 +80,22 @@ function rewriteAllFields(schema, configCallback, context) {
...
+  if (!config.muteWarningTip) {
+    console.warn(
+      'json-schema-form modify(): Make sure you log the returned `warn` as they highlight possible bugs in your modifications. To mute this log, pass `muteWarningTip: true` to the config.'
+    );
+  }
+
+  return {
+    schema,
+    warn: warn.length > 0 ? warn : undefined,
```
eng-almeida, line 99: Wouldn't `warnings` be a more suitable name? warn it's a verb and therefore a function, so it sounds a bit strange to me.

Nitpick: Why not simply return `warn` instead of running a ternary? Or alternatively, if you prefer returning a null value (instead of an empty array) we could initialize `warns` as `null` and set its type later in the code.

danielcardoso5: Agree with the above, I think returning `warnings` at all times – empty if no warnings – would be expected.

I would still go with what I suggested in [this comment](https://github.com/remoteoss/json-schema-form/pull/80#discussion_r1687911026) though. It's non-blocking here since we can add the internal logging later, but it should affect how we name `muteWarningTip` and the existing default warning. If we go with internal logging, I would rename the property to `muteWarnings` or `muteLogging` since it's more generic, and remove the default log.

danielcardoso5: To clarify on why keeping both internal logging and `warnings` array return:

- when using `modify` it's useful to get the warnings right away, no need for devs to do any additional config
- `warnings` return is useful to display these warnings in UI, e.g. json-schema-form own docs

src/modify.js
danielcardoso5: Makes sense @sandrina-p, thanks for splitting this for easy reviewing!!

src/modify.js
sandrina-p: I thought about it but then as in the other PR https://github.com/remoteoss/json-schema-form/pull/80#discussion_r1684195117 was `warn` I let it be. I've renamed now to `warnings`.

But starting as `null` and then change it to array does not work well with mutations. Check this PR and run the tests, you'll see one failing.

Reassigining `warn` (now `warning`) from `null` to `[]` loses the array reference in the original value. and that's why it must be started as array. The alternative is to not use mutation. Instead return the warnings from each one of the mutating functions, like this:

```js
function modify(schema, config) {
        const splitResult = splitFields(schema, configOrder)
        const orderResult = reorderFields(schema, configOrder)
        // ... the same for all the other field mutations, create, remove, etc...

        const warns = [...splitResult.warn, ...orderResult.warn]
        
        return { 
          warn: warns.length > 0 ? warns : null
        }
}
```

After writing this I think I prefer the composable approach instead of mutation 😅

sandrina-p: Give me 2min, I'll show you the example in action right in this MR with the `rewriteFields` missing warning.

danielcardoso5: Why do we want to return `warnings` as null though? Why not always return it and if no warnings it'll be an empty array ?

sandrina-p: Ups, took me longer, as I had to fight recursive warnings. Has tests too. Check it out! https://github.com/remoteoss/json-schema-form/pull/83/commits/42b31ca90ede3c4686f3d1effc45dcc0e234f244

src/tests/modify.test.js
```diff
@@ -171,6 +208,44 @@ describe('modify() - basic mutations', () => {
     });
   });
 
+  it('replace fields that dont exist gets ignored', () => {
+    // IMPORTANT NOTE on this behavior:
+    // Context: At Remote we have a lot of global customization that run equally across multiple different JSON Schemas.
+    // With this, we avoid applying customizations to non-existing fields. (aka create fields)
+    // That's why we have the "create" config, specific to create new fields.
```
sandrina-p, line 215: Note: The create config will come in the other PR https://github.com/remoteoss/json-schema-form/pull/80

src/modify.js
```diff
@@ -80,10 +94,21 @@ function rewriteAllFields(schema, configCallback, context) {
...
   rewriteAllFields(schema, config.allFields);
 
-  return schema;
+  if (!config.muteWarningTip) {
+    console.warn(
+      'json-schema-form modify(): We highly recommend you to handle/report the returned `warnings` as they highlight possible bugs in your modifications. To mute this log, pass `muteWarningTip: true` to the config.'
+    );
+  }
+
+  const allWarnings = [resultRewrite?.warnings].filter(Boolean).flat();
```
danielcardoso5, line 108: You need to flatten before the filter. Also no need for the null check since `resultRewrite` always returns an object

src/modify.js
sandrina-p: @danielcardoso5 so the devs can do `if (!warnings)` without nasty surprises as an empty array being considered truthy 🫠

danielcardoso5: > nasty surprises as an empty array being considered truthy

Why would this be bad? The name suggests an array.

@sandrina-p could you address [these comments above](https://github.com/remoteoss/json-schema-form/pull/83#discussion_r1689581954)? Especially the naming `muteWarningTip` since it affects a prop name that might change.

src/modify.js
```diff
@@ -1,6 +1,9 @@
 import get from 'lodash/get';
 import mergeWith from 'lodash/mergeWith';
 
+const WARNING_TYPES = {
+  CHANGE_FIELD_NOT_FOUND: 'CHANGE_FIELD_NOT_FOUND',
```
danielcardoso5, line 5:
```suggestion
  CHANGED_FIELD_NOT_FOUND: 'CHANGED_FIELD_NOT_FOUND',
```

src/modify.js
sandrina-p: @danielcardoso5 nice catch on the flat to be before, but the `warnings` can be null if the config does not exist (line 31). But I changed the line to return it. Check [b68152d](https://github.com/remoteoss/json-schema-form/pull/83/commits/b68152df3be6f53106d9058f24a90bed994e26c3)

src/modify.js
sandrina-p: 🤔  CHANGED assumes it was indeed changed. CHANGE is the action (change field), which actually did not happen because it was not found.

danielcardoso5: I see, maybe `FIELD_TO_CHANGE_NOT_FOUND`? 😅

danielcardoso5: no need to change, I just find "change field" (action) and "not found" a bit weird. It's the "field" that wasn't found, not the "change field" action

sandrina-p: there you go :D

src/modify.js
sandrina-p: > Why would this be bad? The name suggests an array.

Because based on experience, devs forget about this simple thing. But no strong opinion, so yeah, empty array it is. changed.

danielcardoso5: I see, although I think here it's quite safe. I expect `warnings` to only be used to either log existing warning messages (`warnings.forEach`) – if we don't add internal logging – or to display the warnings in UI ( `warnings.map`), empty array would be okay in both occasions IMO.

I think we just need an updated PR description and we're good to go! 🚀 I'll approve

sandrina-p: Yes, it's minimal. PR description updated :D

PR #87, opened 2024-07-30
fix(file): Validate the value is an instance of File

ricardojoaomestre:
**Description**
When a field of type FILE is running validations (`fileFormat` for example) it is assuming the value is actually a file. If, for some reason it's not an instance of a File, the validation breaks unexpectedly.

**Problem**
A form with a field of type FILE that would allow 1 or more files to be added, due to some bug, is transforming the field value into an object (`{path: "filename.ext"}`). When running validations like `fileFormat` it is throwing an exception and does not continue to validate the rest of the form. The exception happens because in the validation itself, the code is accessing properties in an optimistic way assuming the value is an instance of a file

**Solution**
Add a validation when the field is of type FILE to check if it's actually a File (or array of Files). Use same function to protect other file validations and avoid unexpected errors.

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -2580,7 +2580,7 @@ describe('createHeadlessForm', () => {
       );
       fields = result.fields;
     });
-    describe('and file is of inccorrect format', () => {
+    describe('and file is of incorrect format', () => {
       const file = new File(['foo'], 'file.txt', {
         type: 'text/plain',
       });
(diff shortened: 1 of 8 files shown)


src/tests/createHeadlessForm.test.js
```diff
@@ -2580,7 +2580,7 @@ describe('createHeadlessForm', () => {
       );
       fields = result.fields;
     });
-    describe('and file is of inccorrect format', () => {
+    describe('and file is of incorrect format', () => {
```
sandrina-p, line 2583: @ollyd could you please review this? I was @ricardojoaomestre's buddy here, so we need an unbiased reviewer :)

ollyd: Hey @sandrina-p I didn't get a chance today - this will be the first thing I look at in the morning.

ollyd (approved):

ollyd:
Hey @ricardojoaomestre thanks for this! I've tested your branch locally in Dragon and the validation is working well. Can you please create a dev release and corresponding Dragon MR?

PR #89, opened 2024-08-16
fix(select/radio): Add support for options with boolean type

remotecom:
Description
Adding validation for boolean values in Radio input type.

[Internal ticket](https://linear.app/remote/issue/CXP-742)

Problem
Currently the validation for Radio inputs was set up for either `string` or `number` types. When using `boolean` values the validation defaulted to `string` and errored even if it was correct.

Solution
Adding a validation option for boolean types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.3-dev.20240816092154",
+  "version": "0.11.3-dev.20240819123216",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.2-beta.0",
+  "version": "0.11.3-dev.20240816092154",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.3-dev.20240819123216",
+  "version": "0.11.2-beta.0",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -783,6 +783,8 @@ describe('createHeadlessForm', () => {
     it('support "radio" field boolean type', () => {
       const { fields, handleValidation } = createHeadlessForm(schemaInputTypeRadioBoolean);
 
+      const validateForm = (vals) => friendlyError(handleValidation(vals));
+
       expect(fields).toMatchObject([
         {
           description: 'Are you over 18 years old?',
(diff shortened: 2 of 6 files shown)


src/tests/createHeadlessForm.test.js
```diff
@@ -776,6 +780,74 @@ describe('createHeadlessForm', () => {
...
+
+      assertOptionsAllowed({
+        handleValidation,
+        fieldName: 'over_18',
+        validOptions: [true, false],
+        type: 'boolean',
+      });
+    });
+
+    it('support "radio" field number type', () => {
```
remotecom, line 815: Added test for `number` type, since it was missing

sandrina-p: Perfect @Arcombe-Remote, thanks!

src/tests/createHeadlessForm.test.js
```diff
@@ -776,6 +780,74 @@ describe('createHeadlessForm', () => {
...
+          schema: expect.any(Object),
+          type: 'radio',
+        },
+      ]);
+
+      assertOptionsAllowed({
+        handleValidation,
+        fieldName: 'over_18',
+        validOptions: [true, false],
+        type: 'boolean',
```
sandrina-p, line 811: suggestion: Would you agree this is better?
```suggestion
        type: schemaInputTypeRadioBoolean.properties.over_18.type,
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.2-beta.0",
+  "version": "0.11.3-dev.20240816092154",
```
sandrina-p, line 3: This seems good to go! Let's just wait for the internal CI/CD to be green and then we can revert this dev version, merge this and release the final version!

src/tests/createHeadlessForm.test.js
```diff
@@ -776,6 +780,74 @@ describe('createHeadlessForm', () => {
...
+          required: true,
+          schema: expect.any(Object),
+          type: 'radio',
+        },
+      ]);
+
+      assertOptionsAllowed({
+        handleValidation,
+        fieldName: 'over_18',
+        validOptions: [true, false],
```
sandrina-p, line 810: bug: Something is still wrong in this code because if I pass `"true"` or `"false"`, it's still valid when it should not be.

For example:

```js
  const validateForm = (vals) => friendlyError(handleValidation(vals));

  expect(validateForm({over_18: 'true'})).toEqual({
    over_18: 'The option "true" is not valid.',
  });
```

Arcombe-Remote: Looks like yup casts 'true' to `true` when using boolean()

Arcombe-Remote: `
      validateForm({ over_18: 'true' });
      validateForm({ over_18: 'false' });
      expect(validateForm({over_18: 'maybe'})).toEqual({
        over_18: 'The option "maybe" is not valid.',
      });
`

Screenshot 2024-08-19 at 21 19 02
Screenshot 2024-08-19 at 21 19 08

sandrina-p: Saw the latest commit, looks good! Release a new dev version with it, and once our internal MR has green tests, then we can merge this and release the final version.

sandrina-p: @Arcombe-Remote , I'm optimistic the internal MR will pass, so feel free to rever the version back to what it is in main (package and package-lock) and I'll aprove this right away :)

sandrina-p: @brennj could you re-approved this MR? It was already reviewed/approved by me, but I had to push a commit to revert to the main version. That way I can unblock @Arcombe-Remote when he starts his day tomorrow.

brennj: course

brennj (approved):

src/tests/createHeadlessForm.test.js
```diff
@@ -8,7 +8,9 @@ import {
   JSONSchemaBuilder,
   schemaInputTypeText,
   schemaInputTypeRadioDeprecated,
-  schemaInputTypeRadio,
+  schemaInputTypeRadioString,
```
sandrina-p, line 11: FYI @Arcombe-Remote , I've renamed the PR title to follow conventional commits :) fix(select/radio): Add support for options with boolean type (#89)

sandrina-p: 🎉 Version `0.11.3-beta.0` published!

PR #91, opened 2024-08-20
fix(validation): Validate null field type correctly; files accept object value with `name` property

danielcardoso5:
Using a schema with a field with `type: "null"` that has no `inputType` that differs from `"hidden"`, resulted in an error when performing the schema validation, which ultimately allowed for invalid payload to be submitted.

Example of a schema that presents the invalid behavior:

```json
{
  "type": "object",
  "properties": {
    "name": { "type": "null", "title": "(Optional) Name" },
    "username": { "type": "string", "title": "Username", "maxLength": 2 }
  },
  "required": ["username"]
}
```

Testing in the playground, without this PR, allows submitting an invalid `username` value and the following error is logged:
image

[Internal ticket](https://linear.app/remote/issue/RMT-911)

**Additional bug fixes**

Includes bug fixes for validations introduced in https://github.com/remoteoss/json-schema-form/pull/87. We got internal pipeline failures due to incorrect file field validations. It no longer forces validations on file objects to be of `File` instance.

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -1598,7 +1598,7 @@ describe('createHeadlessForm', () => {
           label: 'Username',
           type: 'text',
           jsonType: 'string',
-          maxLength: 2,
+          maxLength: 4,
           schema: expect.any(Object),
         },
       ]);
@@ -1609,13 +1609,13 @@ describe('createHeadlessForm', () => {
         username: 'Required field',
       });
 
-      expect(validateForm({ username: 'test' })).toEqual({
-        username: 'Please insert up to 2 characters',
+      expect(validateForm({ username: 'hello' })).toEqual({
+        username: 'Please insert up to 4 characters',
       });
 
-      expect(validateForm({ username: 'hi' })).toBeUndefined();
-      expect(validateForm({ name: 2, username: 'hi' })).toBeUndefined();
-      expect(validateForm({ username: 'hi' })).toBeUndefined();
+      expect(validateForm({ username: 'john' })).toBeUndefined();
+      expect(validateForm({ name: 2, username: 'john' })).toBeUndefined();
+      expect(validateForm({ username: 'john' })).toBeUndefined();
     });
 
     it('supports "null" field type with inputType "hidden"', () => {
@@ -1634,7 +1634,15 @@ describe('createHeadlessForm', () => {
           label: 'Username',
           type: 'text',
           jsonType: 'string',
-          maxLength: 2,
+          maxLength: 4,
+          schema: expect.any(Object),
+        },
+        {
+          name: 'age',
+          label: 'Age',
+          type: 'number',
+          jsonType: 'null',
+          maximum: 20,
           schema: expect.any(Object),
         },
       ]);
@@ -1645,13 +1653,14 @@ describe('createHeadlessForm', () => {
         username: 'Required field',
       });
 
-      expect(validateForm({ username: 'test' })).toEqual({
-        username: 'Please insert up to 2 characters',
+      expect(validateForm({ username: 'hello', age: 25 })).toEqual({
+        username: 'Please insert up to 4 characters',
+        age: 'Must be smaller or equal to 20',
       });
 
-      expect(validateForm({ username: 'hi' })).toBeUndefined();
-      expect(validateForm({ name: 2, username: 'hi' })).toBeUndefined();
-      expect(validateForm({ username: 'hi' })).toBeUndefined();
+      expect(validateForm({ username: 'john', age: 19 })).toBeUndefined();
+      expect(validateForm({ name: 2, username: 'john' })).toBeUndefined();
+      expect(validateForm({ username: 'john' })).toBeUndefined();
     });
 
     it('supports oneOf pattern validation', () => {
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -2874,6 +2874,26 @@ describe('createHeadlessForm', () => {
         ).resolves.toEqual({ fileInput: [{ name: 'foo.pdf' }] });
       });
 
+      it('should validate format', async () =>
+        expect(
+          object()
+            .shape({
+              fileInput: fields[0].schema,
+            })
+            .validate({ fileInput: [{ name: 'foo.txt' }] })
+        ).rejects.toMatchObject({
+          errors: ['Unsupported file format. The acceptable formats are .png,.jpg,.jpeg,.pdf.'],
+        }));
+
+      it('should validate max size', async () =>
+        expect(
+          object()
+            .shape({
+              fileInput: fields[0].schema,
+            })
+            .validate({ fileInput: [{ name: 'foo.txt', size: 1024 * 1024 * 1024 }] })
+        ).rejects.toMatchObject({ errors: ['File size too large. The limit is 20 MB.'] }));
+
       it('throw an error if invalid file object', async () =>
         expect(
           object()
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -29,6 +29,7 @@ import {
   schemaInputWithStatement,
   schemaInputTypeCheckbox,
   schemaInputTypeCheckboxBooleans,
+  schemaInputTypeNull,
   schemaWithOrderKeyword,
   schemaWithPositionDeprecated,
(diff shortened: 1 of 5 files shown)


src/tests/createHeadlessForm.test.js
```diff
@@ -1578,6 +1580,89 @@ describe('createHeadlessForm', () => {
...
+      expect(validateForm({})).toEqual({
+        username: 'Required field',
+      });
+
+      expect(validateForm({ username: 'hello' })).toEqual({
+        username: 'Please insert up to 4 characters',
+      });
+
+      expect(validateForm({ username: 'john' })).toBeUndefined();
+      expect(validateForm({ name: 2, username: 'john' })).toBeUndefined();
```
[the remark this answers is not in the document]
sandrina-p: If you are curious, check jsonschemalint.com, it fails there 

<details>

image
<details>

src/tests/helpers.js
```diff
@@ -2213,3 +2213,51 @@ export const schemaWithCustomValidationsAndConditionals = {
...
+      'x-jsf-presentation': {
+        inputType: 'text',
+      },
+    },
+    age: {
+      type: 'null',
+      title: 'Age',
+      maximum: 20,
+      'x-jsf-presentation': {
+        inputType: 'number',
```
sandrina-p, line 2258: question/bug: How is this scenario possible? Having a field of `inputType: number` but with `type null`?

src/tests/createHeadlessForm.test.js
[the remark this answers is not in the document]
danielcardoso5: hum, maybe I'm confusing the schema – and its validations – with the presentational aspect of a `type: null` field. From the docs, a `type: null` defaults to a text input, and it currently allows submitting the value of that text input, probably incorrectly though. Thanks for sharing the lint example, I'll update to only accept `null`/absent values.

src/tests/helpers.js
danielcardoso5: I'm pretty sure this is just confusion in my head of "json type" vs "input type" 😅 will clean this up!

src/tests/createHeadlessForm.test.js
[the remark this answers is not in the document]
sandrina-p: Good catch, so that's a bug or miss-understanding... That made sense until now, but now that we are considering fields that don't carry any value, then it doesn't make sense. We can keep the default to `text` (as we don't have the concept of `hidden` in JSF). But the validation must be fixed to now allow it (that's a bug) 🐛  WDYT?

src/tests/helpers.js
sandrina-p: Feel free to clarify [the docs](https://json-schema-form.vercel.app/?path=/docs/guides-concepts-important-concepts--docs) on what's the difference! Basically jsonType is native specs, `inputType` is opinionated UI visualization of that same field. For example a `jsonType: string` can map to multiple `inputType`: `text`, `radio`, `select`, etc..

src/tests/createHeadlessForm.test.js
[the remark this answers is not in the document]
danielcardoso5: Maybe that's a documentation issue based on a leak of a Dragon abstraction (DynamicForm) into json-schema-form docs?

The output of this field has no `inputType` attribute contrary to what the mapping in the docs suggests:
image

So in a way, JSF is not opinionated on what input type should be used for `null` types. Maybe the docs should look like this?

<details><summary>Screenshot</summary>
<p>

image


</p>
</details> 

> But the validation must be fixed to now allow it (that's a bug) 🐛 WDYT?

Done!

src/tests/helpers.js
danielcardoso5: Will add some clarification to the docs along the same lines of what you shared. Thank you!

src/tests/createHeadlessForm.test.js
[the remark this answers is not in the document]
sandrina-p: Then even better, update the docs for that.

src/tests/createHeadlessForm.test.js
```diff
@@ -1578,6 +1579,44 @@ describe('createHeadlessForm', () => {
...
+      const { handleValidation, fields } = createHeadlessForm(schemaInputTypeNull, {
+        strictInputType: false,
+      });
+
+      expect(fields).toMatchObject([
+        {
+          name: 'name',
+          label: '(Optional) Name',
+          type: undefined,
+          jsonType: 'null',
```
sandrina-p, line 1592: Let's assert here that `inputType` is undefined based on your last discovery ^.^

sandrina-p: In case you are wondering, `type` is tech-debt/deprecated. It should not exist (but we still use it a lot in our internal codebases). It's either `inputType` or `jsonType`.

src/yupSchema.js
```diff
@@ -213,7 +220,7 @@ const yupSchemasToJsonTypes = {
   object: yupSchemas.fieldset,
   array: yupSchemas.multiple.select,
   boolean: yupSchemas.checkboxBool,
-  null: noop,
+  null: yupSchemas.null,
```
sandrina-p, line 223: todo: Pls create a dev release for us to run the internals tests. Once green, then we can approve/merge this :D

src/tests/createHeadlessForm.test.js
danielcardoso5: Good idea! Done!

src/yupSchema.js
danielcardoso5: On it!

src/yupSchema.js
```diff
@@ -469,14 +469,16 @@ export function buildYupSchema(field, config, logic) {
...
     validators[0] = () => withBaseSchema().shape(buildFieldSetSchema(propertyFields.fields));
-  } else if (inputType === supportedTypes.FILE) {
-    validators.push(withFile);
   }
 
   if (propertyFields.required) {
     validators.push(withRequired);
   }
 
+  if (inputType === supportedTypes.FILE) {
```
sandrina-p, line 478: thought: Good catch. I never thought the order of validators would matter here 💡

danielcardoso5: Indeed, took me a while to catch what was going on!

danielcardoso5:
@sandrina-p I've added the schema for the upcoming docs in https://github.com/remoteoss/json-schema-form/pull/91/commits/623c31b5c5d2aedfb1e6ad7f070e3453909cb878

It doesn't contain the `ignoreValue` config since that's an internal concept and doesn't change the behavior of JSF.

sandrina-p (approved):

src/tests/createHeadlessForm.test.js
```diff
@@ -29,6 +29,7 @@ import {
   schemaInputWithStatement,
   schemaInputTypeCheckbox,
   schemaInputTypeCheckboxBooleans,
+  schemaInputTypeNull,
```
sandrina-p, line 32: @danielcardoso5 approved, let's go! We need to change the PR title though

src/tests/helpers.js
```diff
@@ -782,6 +782,25 @@ export const schemaInputWithStatement = {
   },
 };
 
+export const schemaHiddenInputWithStatement = {
```
sandrina-p, line 785: > @sandrina-p I've added the schema for the upcoming docs in https://github.com/remoteoss/json-schema-form/commit/623c31b5c5d2aedfb1e6ad7f070e3453909cb878

@danielcardoso5 I want your opinion here. Do you agree with this pattern of creating a schema but not used by the codebase directly? I find it very odd... I'd rather simply write in our internal codebase. If it gets stale/invalid the tests will fail anyway.

Context: When we open-sourced JSF we didn't spend much time in this file because of speed/velocity. And now we are paying the bill (engineers following a pattern that doesn't make sense.)

sandrina-p: To clarify, I don't expect you to fix/find-the-perfect-solution right now. Just want your opinion, and then we can tackle in later on.

danielcardoso5: @sandrina-p yeah it's odd, this is just a symptom of the documentation leaving separately from this repo, if we moved it over this wouldn't be a problem, but you already know this 😄 

We could move the ones used for the documentation to our internal codebase, one immediate advantage is that you don't need to do anything in this repo if you just want to improve the docs by adding or changing an example. In this case I added the schema right away to avoid creating another PR just for it.

PR #93, opened 2024-10-09
fix(modify): Support json schemas without allOf

tcrammond:
## What this PR does

Fix an error that occurs if you pick fields using `modify()` on a schema that doesn't define `allOf`.

[Internal ticket](https://linear.app/remote/issue/INF-639)

### Why it happened / how it's fixed

Unguarded access to `allOf.forEach(...)` on the modified schema, which we now check.

### How to reproduce and test

The following example no longer errors on this branch. The added test case demonstrates this.

```js
const schema = {
 // This schema does not use `allOf`
  properties: {
    age: {
      title: 'Age',
      type: 'integer',
    },
    quantity: {
      title: 'Quantity',
      type: 'integer',
    },
  },
  'x-jsf-order': ['age', 'quantity'],
};

// TypeError reading properties of allOf
modify(schema, {
  pick: 'age'
})
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ npm test
 
 Submit your branch pointing to `main`.
 
-Please, always add tests to your bug fixs and new features.
+Please, always add tests to your bug fixes and new features.
 
 ### Testing the PR in your project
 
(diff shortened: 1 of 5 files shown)


ollyd:
Thanks @tcrammond this looks good, do you have a corresponding Dragon MR?

tcrammond:
> Thanks @tcrammond this looks good, do you have a corresponding Dragon MR?

Hi @ollyd, [internal ticket here](https://linear.app/remote/issue/INF-639).

ollyd (approved):

PR #94, opened 2024-10-25
fix(validation): validate radio with string-y types;

dilvane:
This PR updates the validation for a radio input that was incorrectly accepting a boolean, even though the type was defined as a string.

```json
{
  "type": "object",
  "properties": {
    "has_pet": {
      "title": "Has Pet",
      "description": "Do you have a pet?",
      "oneOf": [
        {
          "title": "Yes",
          "const": "yes"
        },
        {
          "title": "No",
          "const": "false"
        }
      ],
      "x-jsf-presentation": {
        "inputType": "radio"
      },
      "type": "string"
    }
  },
  "required": [
    "has_pet"
  ]
}
```

| Before | After |
| ------ | ------ |
| #before|#after|

### DIFF
--- a/src/tests/createHeadlessForm.test.js
+++ b/src/tests/createHeadlessForm.test.js
@@ -9,6 +9,7 @@ import {
   schemaInputTypeText,
   schemaInputTypeRadioDeprecated,
   schemaInputTypeRadioString,
+  schemaInputTypeRadioStringY,
   schemaInputTypeRadioBoolean,
   schemaInputTypeRadioNumber,
   schemaInputTypeRadioRequiredAndOptional,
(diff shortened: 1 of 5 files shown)


src/tests/createHeadlessForm.test.js
```diff
@@ -9,6 +9,7 @@ import {
   schemaInputTypeText,
   schemaInputTypeRadioDeprecated,
   schemaInputTypeRadioString,
+  schemaInputTypeRadioStringY,
```
sandrina-p, line 12: Hey Dilvane! I won't be able to review this as I'm catching up from PTO. @brennj instead can you cover me here?

src/tests/createHeadlessForm.test.js
```diff
@@ -820,6 +821,45 @@ describe('createHeadlessForm', () => {
       });
     });
 
+    // @BUG COD-1859
+    // it should validate when type is string but value is not a boolean
+    it('support "radio" field string-y type', () => {
```
ollyd, line 826: **Suggestion** 

Not sure this needs a separate test case.
We could just ensure the string test case on line 744 ensures a string by updating it to use:
```
          options: [
            {
              label: 'Yes',
              value: 'true',
            },
            {
              label: 'No',
              value: 'false',
            },
          ]
```
Then we can just add this there:
```
      expect(validateForm({ has_siblings: false })).toEqual({
        has_siblings: 'The option "false" is not valid.',
      });
```

ollyd:
Hey @dilvane thanks for the MR! Your second (After) screenshot was a little confusing at first because the schema is correct but there's a validation error, but I remembered it's due to how we treat true/false strings in Dragon. 

Do you have a corresponding Dragon MR to ensure everything is green?

dilvane:
> Hey @dilvane thanks for the MR! I've tested it locally and it's working as expected. Your second (After) screenshot was a little confusing at first because the schema is correct but there's a validation error, but I remembered it's due to how we treat true/false strings in Dragon.
> 

Thanks for checking it @ollyd. 
You're right, probably also because of the error message we display.

The option false is not valid. <- When the schema accepts a string but receives a boolean instead.
The option "false" is not valid. <- When the schema accepts a boolean but receives a string instead.

> Do you have a corresponding Dragon MR to ensure everything is green? 

Yes, here https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/32817

ollyd (approved):

PR #95, opened 2024-10-31
Add test for conditional to show/hide dependent values

RuslanZavacky:
Thank you for the library! I am experimenting with it, and noticed that example at https://json-schema-form.vercel.app/?path=/story/demos-combinations--conditional-fields doesn't work when this library is used independently.

conditional_json_schema

I'd expect when calling handleValidation, for the `fields` object to get isVisible updated correctly. 

I understand that `formValuesToJsonValues` is used to remove fields that are in conflict with conditionals, but the result of that function is not the one you would use to render form. I might be missing something there.

I've added test that shows that field isVisible property is in unexpected state.

----

To render form elements, I rely on `fields` and isVisible. And I assume it is not the same logic that is used to build out demo, as it is doing something else to derive, if field is visible or not.

### DIFF
--- a/src/tests/conditions.test.js
+++ b/src/tests/conditions.test.js
@@ -479,3 +479,140 @@ describe('Conditional with a minimum value check', () => {
     expect(handleValidation({ salary: 1000, reason: 'reason_one' }).formErrors).toEqual(undefined);
   });
 });
+
+describe('Conditional with fields that depend on the previous value', () => {
+  it('Should hide dependent field if value does not satisfy conditional anymore', () => {
+    const schema = {
+      type: 'object',
+      additionalProperties: false,
+      properties: {
+        has_pet: {
+          title: 'Has Pet',
+          description: 'Do you have a pet?',
+          oneOf: [
+            {
+              title: 'Yes',
+              const: 'yes',
+            },
+            {
+              title: 'No',
+              const: 'no',
+            },
+          ],
+          'x-jsf-presentation': {
+            inputType: 'radio',
+          },
+          type: 'string',
+        },
+        pet_name: {
+          title: "Pet's name",
+          description: "What's your pet's name?",
+          'x-jsf-presentation': {
+            inputType: 'text',
+          },
+          type: 'string',
+        },
+        pet_age: {
+          title: "Pet's age",
+          description: "What's your pet's age",
+          'x-jsf-presentation': {
+            inputType: 'number',
+          },
+          type: 'number',
+        },
+        dietary_needs: {
+          title: 'Dietary needs',
+          description: "What are your pet's dietary needs?",
+          'x-jsf-presentation': {
+            inputType: 'textarea',
+          },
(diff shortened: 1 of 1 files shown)


src/tests/conditions.test.js
```diff
@@ -479,3 +479,140 @@ describe('Conditional with a minimum value check', () => {
     expect(handleValidation({ salary: 1000, reason: 'reason_one' }).formErrors).toEqual(undefined);
   });
 });
+
+describe('Conditional with fields that depend on the previous value', () => {
```
sandrina-p, line 483: Hi @RuslanZavacky , thank you a lot for this bug report and tests! Someone in our team will look into it later in a couple of days max 👀

sandrina-p: Hi @RuslanZavacky! I had to the time now to dig into this. Explaining it is tricky, so I pushed a new commit with the data flow explanation, and alternatives. Look for the 📌 emoji in the code!

Spoiler: I did not fix the bug but provided alternatives. Let me know if it makes sense.

sandrina-p: P.S. I amended the commit, here's the new version: [d7f80b9](https://github.com/remoteoss/json-schema-form/pull/95/commits/d7f80b964e940b8ff8f353e164e3866fa0ccd5fa)

sandrina-p: No, we don't plan to look into it in the next few days (or weeks). If you could give it a try, we'd highly appreciate. :)

But be aware that this problem doesn't get fixed by just doing a re-calculation. In this case we only have 1 condition that depends on another condition. (a chain of 2 conditions), so a re-calculation works. But given a JSON Schema with a chain of 3, 4, 5+ conditions, it means we'd need to do the re-calculation 3,4,5+ times.

I think the best solution is fixing the JSON Schema condition instead to account for the 2 dependent fields (has_pet and age) as shown in the unit test I created. Does that make sense to you?

RuslanZavacky: Yes, that makes sense :) Thank you!

RuslanZavacky:
@sandrina-p do you think it is worth to merge this test case, or to cancel this PR? I feel like trying to solve the underlying recomputation for validation will need a separate ticket/pr, as it looks like quite a lot of effort might be needed :)

src/tests/conditions.test.js
```diff
@@ -479,3 +509,348 @@ describe('Conditional with a minimum value check', () => {
...
+  it('Should hide dependent field if value does not satisfy conditional anymore', () => {
```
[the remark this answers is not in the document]
sandrina-p: Hi @RuslanZavacky I confess I forgot this PR and then went on vacation. In the meantime, the playground is already using the "chained conditionals" as explained above to avoid the original bug from happening.

As you didn't answer, I'm closing this PR. But if you want to cleanup the test with that explanation, go ahead, we'll merge it after :)

PR #98, opened 2024-11-27
chore(json-logic): Add skipped broken unit test for missing json-logic `reduce` support

brennj:
This was a bug brought up internally when trying to create a schema similar to the test case. This is because of a false assumption on my side where I didn't need to account for "temp variables" that are indeed supported in json logic for cases like `reduce`.

### DIFF
--- a/src/jsonLogic.js
+++ b/src/jsonLogic.js
@@ -399,7 +399,6 @@ function checkRuleIntegrity(
     subRule.map((item) => {
       const isVar = item !== null && typeof item === 'object' && Object.hasOwn(item, 'var');
       if (isVar) {
-        if (Array.isArray(item.var)) return; // Is an accumlator for reduce and can't be checked.
         const exists = jsonLogic.apply({ var: removeIndicesFromPath(item.var) }, data);
         if (exists === null) {
           throw Error(errorMessage(item));
--- a/src/jsonLogic.js
+++ b/src/jsonLogic.js
@@ -399,6 +399,7 @@ function checkRuleIntegrity(
     subRule.map((item) => {
       const isVar = item !== null && typeof item === 'object' && Object.hasOwn(item, 'var');
       if (isVar) {
+        if (Array.isArray(item.var)) return; // Is an accumlator for reduce and can't be checked.
         const exists = jsonLogic.apply({ var: removeIndicesFromPath(item.var) }, data);
         if (exists === null) {
           throw Error(errorMessage(item));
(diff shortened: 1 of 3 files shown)


src/jsonLogic.js
```diff
@@ -399,6 +399,7 @@ function checkRuleIntegrity(
     subRule.map((item) => {
       const isVar = item !== null && typeof item === 'object' && Object.hasOwn(item, 'var');
       if (isVar) {
+        if (Array.isArray(item.var)) return; // Is an accumlator for reduce and can't be checked.
```
ollyd, line 402: **Suggestion**

Just wondering if there could ever be an array path which we would want to validate? 🤔 
Might be safer to be less permissive: `const isReducerVar = operator === 'reduce' && Array.isArray(item.var)`
This also makes the code more explicit so we don't need code comments. 🙂

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.8-beta.0",
+  "version": "0.11.9-dev.20241127210659",
```
sandrina-p, line 3: Hi! Forgive me, I can't review this, I has been wrapping up my work before going on vacation for the next two weeks. I'll defer to @ollyd here :)

sandrina-p: @brennj can what's the status of this PR? If I remember correctly, you were not able to fully fix the bug.

If that's true, could I ask you a favor. Could you update the MR to only have a unit test describing the bug, so we make sure to fix it in the upcoming `next/` version. Similar to #107 thanks!

brennj: finally getting around to this, i'm gonna add the test case as a todo as you mention ✍️

brennj:
rebased with `main` @dragidavid / @thien-remote with the move to `next` I'll leave this in your capable hands to ensure this is eventually moved into there. Feel free to merge/disregard this PR!

src/tests/jsonLogic.test.js
```diff
@@ -243,6 +244,23 @@ describe('jsonLogic: cross-values validations', () => {
     });
   });
 
+  // TODO: Implement this test.
+  describe.skip('Reduce', () => {
```
sandrina-p, line 248: > rebased with main @dragidavid / @thien-remote with the move to next I'll leave this in your capable hands to ensure this is eventually moved into there. Feel free to merge/disregard this PR!

I was checking this now, I bet it got missed by David or Thien. All good, I'm defaulting to action in merging this now (as it's only a new test case, nothing changed.) The team will take care of it as part of [this internal issue](https://linear.app/remote/issue/DEVXP-2547)

cc @lukad

sandrina-p (approved):

PR #100, opened 2024-12-04
fix: Add support for boolean values in the if condition

waiting-for-dev:


src/tests/internals.helpers.test.js
```diff
@@ -173,3 +173,21 @@ describe('pickXKey()', () => {
     expect(xKey).toBeUndefined();
   });
 });
+
+describe('processNode()', () => {
```
brennj, line 177: im not a big fan of testing the internals, i dont like this file in general.

I think what could be better would be two schemas, one like you have and then a `true` case in the `conditions.test.js` file, just testing it from the top level API perspective

Other than that, this change works great

waiting-for-dev: Thanks for the review @brennj. It makes total sense! I updated the code. Please, let me know what you think 🙂

src/tests/conditions.test.js
```diff
@@ -479,3 +479,71 @@ describe('Conditional with a minimum value check', () => {
...
+    expect(fields[0]).toMatchObject({
+      name: 'is_full_time',
+      required: false,
+    });
+    expect(fields[1]).toMatchObject({
+      name: 'salary',
+      required: true,
+    });
+  });
+});
```
brennj, line 549: these are much much better, thank you for adding

brennj (approved):
All good here my side @waiting-for-dev but please open an internal MR first and make sure it passes on our FE codebase before merging

waiting-for-dev:
> All good here my side @waiting-for-dev but please open an internal MR first and make sure it passes on our FE codebase before merging

If I understand it well, do I need to [ask you for a dev release](https://github.com/remoteoss/json-schema-form/blob/main/CONTRIBUTING.md#public-release)? 😄

waiting-for-dev:
> > All good here my side @waiting-for-dev but please open an internal MR first and make sure it passes on our FE codebase before merging
> 
> If I understand it well, do I need to [ask you for a dev release](https://github.com/remoteoss/json-schema-form/blob/main/CONTRIBUTING.md#public-release)? 😄

Oh, I now see any Remoter should be able to do it. Gonna try!

waiting-for-dev:
> All good here my side @waiting-for-dev but please open an internal MR first and make sure it passes on our FE codebase before merging

@brennj if I'm not wrong you need to add me in the GH repo. When running `npm run release:dev:patch`:

> ERROR: Permission to remoteoss/json-schema-form.git denied to waiting-for-dev.
> fatal: Could not read from remote repository.

PR #103, opened 2024-12-10
Multiple conditions affecting the same field are respected

brennj:
Does this by adding `dynamicAttributes` when computed conditional properties to ensure they all apply. 

* Still has some quirks that don't fully respect the JSON Schema spec. But this is a huge win for us to be able to have multiple allOf conditions respected.

### DIFF
--- a/src/tests/conditions.test.js
+++ b/src/tests/conditions.test.js
@@ -511,4 +511,85 @@ describe('Multiple conditions affecting the same field', () => {
     expect(handleValidation({ answer: 'no', number: 15 }).formErrors).toBeUndefined();
     expect(handleValidation({ answer: 'no', number: 21 }).formErrors).toBeUndefined();
   });
+
+  it('Accepts unsatisfiable conflicting conditions', () => {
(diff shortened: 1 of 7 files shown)


src/tests/conditions.test.js
```diff
@@ -479,3 +479,169 @@ describe('Conditional with a minimum value check', () => {
...
+    });
+    expect(handleValidation({ answer: 'yes', number: 21 }).formErrors).toEqual({
+      number: 'Must be smaller or equal to 20',
+    });
+    expect(handleValidation({ answer: 'no', number: 9 }).formErrors).toBeUndefined();
+    expect(handleValidation({ answer: 'no', number: 15 }).formErrors).toBeUndefined();
+    expect(handleValidation({ answer: 'no', number: 21 }).formErrors).toBeUndefined();
+  });
+
+  it('Should apply all conditions - same keyword', () => {
```
sandrina-p, line 521: bug: I added this test showing the pending bug

PR #105, opened 2025-01-09
chore(tests): Slightly better tests organization

sandrina-p:
As the name says, I slightly improved the tests organization. I time-boxed ~30m for this. What I did:

- Added new `describe()`s.
- Marked which tests are about deprecated features (not meant to be in v2)
- Some minor reorders.
- Migrated some `JSONSchemaBuilder` to normal JSON Schemas, but not all, it was more than 50. I'll leave that for another MR.

What I did not do:
- Refactors to tests logic, otherwise it's a mess.
- I did not create new files, it would be harder to track the changes. 

Yet, even with these efforts, the code review is barely impossible. How can you trust me?

The nr of tests between `main` and this PR is 1 less test. (now it's 278) I deleted the test `pass "description" to field` because it was redundant compared to the tests `supports both root level "description"`. Everything else is the same.

```
Test Suites: 9 passed, 9 total
Tests:       2 todo, 276 passed, 278 total
Snapshots:   2 passed, 2 total
```

My goal at this point was just making the tests tree a little bit easier to understand. Here's multiple screenshots of Before VS Now:

image

image

image

image

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -5,7 +5,7 @@ import omitBy from 'lodash/omitBy';
 import set from 'lodash/set';
 import { lazy } from 'yup';
 
-import { checkIfConditionMatchesProperties } from './checkIfConditionMatches';
+import { checkIfConditionMatchesProperties } from './internals/checkIfConditionMatches';
 import { supportedTypes, getInputType } from './internals/fields';
 import { pickXKey } from './internals/helpers';
 import { processJSONLogicNode } from './jsonLogic';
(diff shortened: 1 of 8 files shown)


src/helpers.js
```diff
@@ -5,7 +5,7 @@ import omitBy from 'lodash/omitBy';
 import set from 'lodash/set';
 import { lazy } from 'yup';
```
sandrina-p, line 6: @antoniocapelo why can't I tag you as reviewer? 👀

ollyd (approved):
Hey @sandrina-p I saw the notifications for these changes so even though I wasn't marked as a Reviewer, the changes LGTM so I'm approving just in case you want to go ahead and get this merged in.

PR #107, opened 2025-01-16
chore(conditionals): Highlight bugs related to validations and conditional fields 

sandrina-p:
I've added a few tests highlighting existing bugs in JSF. For awareness and to make sure they get solved when the next major version is developed:

* c525914 - multiple conditionals to the same field are not applied 
* 0403f72 - values passed to handleValidation are mutated, but should not 
* e232aba - conditional hidden fields should throw an error, but they do not

[Internal ticket](https://linear.app/remote/issue/DEVXP-2535)

### DIFF
--- a/src/tests/conditions.test.js
+++ b/src/tests/conditions.test.js
@@ -626,4 +626,20 @@ describe('Conditionals - bugs and code-smells', () => {
     // The error should be something like:
     // expect(validation2.formErrors).toEqual({ pet_name: 'Not allowed.'});
   });
+
+  it('Given values from hidden fields, it mutates the values (bug behavior)', () => {
+    const { handleValidation } = createHeadlessForm(schemaHasPet);
+
+    const newValues = { has_pet: 'no', pet_name: 'Max' };
+
+    const validation = handleValidation(newValues);
+
+    expect(newValues).toEqual({
+      has_pet: 'no',
+      pet_name: null, // BUG! 🐛 Should still be "Max", should not be mutated.
+    });
+
+    // Same bug as explained in the previous test.
+    expect(validation.formErrors).toBeUndefined();
+  });
 });
--- a/src/tests/conditions.test.js
+++ b/src/tests/conditions.test.js
@@ -642,4 +642,86 @@ describe('Conditionals - bugs and code-smells', () => {
     // Same bug as explained in the previous test.
     expect(validation.formErrors).toBeUndefined();
   });
+
+  it('Given multiple conditionals to the same field, it only applies the last one (bug behavior) - case 1', () => {
+    const { handleValidation } = createHeadlessForm(
+      {
+        additionalProperties: false,
+        properties: {
+          field_a: { type: 'string' },
+          field_b: { type: 'string' },
+          field_c: { type: 'number' },
+        },
+        allOf: [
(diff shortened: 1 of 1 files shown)


src/tests/conditions.test.js
```diff
@@ -547,3 +547,170 @@ describe('Conditional with literal booleans', () => {
...
+      strictInputType: false,
+    });
+
+    const petNameField = fields[1];
+
+    const validation1 = handleValidation({ has_pet: 'yes' });
+    expect(petNameField.isVisible).toBe(true);
+    expect(validation1.formErrors).toEqual({
+      pet_name: 'Required field',
+    });
```
antoniocapelo, line 606: **Question**

Do we need these lines in the test? I.e., does the test still fail if we remove them. If so, we should probably add it as a comment at the top of the test. If not, maybe we should remove them

antoniocapelo (approved):

src/tests/conditions.test.js
sandrina-p: Good point. No, the test doesn't fail if we remove them. It's just for sanity check that other basic behavior still works. Would you still remove it?

src/tests/conditions.test.js
```diff
@@ -547,3 +547,170 @@ describe('Conditional with literal booleans', () => {
...
+            pet_name: false,
+          },
+        },
+      },
+    ],
+  };
+
+  it('Given values from hidden fields, it does not thrown an error (bug behavior)', () => {
+    const { fields, handleValidation } = createHeadlessForm(schemaHasPet, {
+      strictInputType: false,
```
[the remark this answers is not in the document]
sandrina-p: Yes it is. `strictInputType: false` allows passing a JSON Schema without 'x-jsf-presentation.inputType'. As that part is irrelevant, I prefered to keep the schema as clean/short as possible.

src/tests/conditions.test.js
sandrina-p: I removed it, simply because we already have plenty of other tests ensuring that normal validations to "Required field" work as expected

src/tests/conditions.test.js
```diff
@@ -547,3 +547,170 @@ describe('Conditional with literal booleans', () => {
...
+
+    const validation1 = handleValidation({ has_pet: 'yes' });
+    expect(petNameField.isVisible).toBe(true);
+    expect(validation1.formErrors).toEqual({
+      pet_name: 'Required field',
+    });
+
+    const validation2 = handleValidation({ has_pet: 'no', pet_name: 'Max' });
+    expect(petNameField.isVisible).toBe(false);
+    // Bug: 🐛 It does not thrown an error,
```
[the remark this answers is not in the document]
sandrina-p: Technically all tests are `@todo v2` (remember, parity with v1!) 😅 But I get your point, so I marked it as @bug to make clear it's an existing bug and should be fixed.

lukad:
> Hey @sandrina-p all's good for me! But looks like both me and Capelo don't have write permission, so we will need someone who do (@lukad ?) or we could join the maintainer list 🤔

Ah, sorry @thien-remote, I missed your comment. I'll have a look at the repo's settings.

PR #112, opened 2025-01-22
build(next): Simplify test setup

thien-remote:
This PR simplifies the test setup in https://github.com/remoteoss/json-schema-form/pull/111, as per @thien-remote 's first review. It is intentionally extracted as another PR to avoid invalidating the review. It should also make it easier to follow.


babel.config.js
dragidavid, line 1: I'm pretty sure (pun intended) that we have different prettier settings since if I just save this file, it brings back the multi line format 😂

next/README.md
dragidavid, line 1: Thanks for adding this

PR #114, opened 2025-01-22
build(next): Add release script 

dragidavid:
To follow up on https://github.com/remoteoss/json-schema-form/pull/108, https://github.com/remoteoss/json-schema-form/pull/110 and https://github.com/remoteoss/json-schema-form/pull/111, this PR adds the release scripts to `next`. It belongs to the same internal ticket of [DEVXP-2530](https://linear.app/remote/issue/DEVXP-2530/set-up-json-schema-form-v2-package).

---

This PR adds the new `release_next` script which publishes the `next` package to NPM tagged with `next` for easy distinction.

The script is based on the `dev` and `main` release scripts.


scripts/release_next.js
```diff
@@ -0,0 +1,98 @@
...
+
+  if (!versionType) {
+    console.log('🟠 version type is missing. Make sure to run the script from package.json');
+    process.exit(1);
+  }
+
+  // Keep alpha tag for next versions
+  const versionBase = semver.coerce(currentVersion);
+  return semver.inc(versionBase, versionType) + '-alpha.0';
+}
```
dragidavid, line 43: Ideally, we'd be bumping the version number after this part.

scripts/release_next.js
```diff
@@ -0,0 +1,98 @@
...
+  console.log('Committing published version...');
+  const cmd = `git add next/package.json && git commit -m "Release next ${newVersion}" && git tag next-v${newVersion} && git push && git push origin --tags`;
+  await runExec(cmd);
+}
+
+async function publish({ newVersion, otp }) {
+  console.log('Publishing new version...');
+
+  // Use --tag=next to mark this as the "next" version
+  const cmd = `cd next && npm publish --access=public --tag=next --otp=${otp}`;
```
dragidavid, line 66: Tagged as `next`

scripts/release_next.js
```diff
@@ -0,0 +1,98 @@
...
+  const versionType = process.argv.slice(2)[0];
+
+  if (!versionType) {
+    console.log('🟠 version type is missing. Make sure to run the script from package.json');
+    process.exit(1);
+  }
+
+  // Keep alpha tag for next versions
+  const versionBase = semver.coerce(currentVersion);
+  return semver.inc(versionBase, versionType) + '-alpha.0';
```
antoniocapelo, line 42: Can't we do something like [the docs](https://github.com/npm/node-semver?tab=readme-ov-file#prerelease-identifiers) mention?

```
  return semver.inc(versionBase, versionType, 'alpha');
```

dragidavid: Lets try that!

scripts/release_next.js
```diff
@@ -0,0 +1,98 @@
...
+    process.exit(1);
+  }
+
+  // Keep alpha tag for next versions
+  const versionBase = semver.coerce(currentVersion);
+  return semver.inc(versionBase, versionType, 'alpha');
+}
+
+async function bumpVersion({ newVersion }) {
+  const cmd = `cd next && npm version --no-git-tag-version ${newVersion}`;
```
antoniocapelo, line 46: **Question**

We're using `--no-git-tag-version ${newVersion}` and then later we're manually commiting the package bump — any reason for that?

What I mean is, if we'd run `npm version ${newVersion}`, couldn't remove the need for the `gitCommit` function?

dragidavid: Great question! To be honest this was taken from the `dev` and `main` release scripts so I didn't think too much of it. Now that I check it might help with the tag and prefix since `npm version` on its own just bumps the version without the prefix. 🤔

antoniocapelo (approved):

dragidavid:
Going to leave this unmerged until the morning, in case we get some more eyes on it overnight. 👀

scripts/release_next.js
sandrina-p: I was the one writing the original scripts. Original PR: https://github.com/remoteoss/json-schema-form/pull/3. Your conclusion is exactly that. We have more control over the commit msg.

package.json
```diff
@@ -37,6 +37,7 @@
     "release:dev:minor": "node scripts/release_dev minor",
     "release:main:patch": "node scripts/release_main patch",
     "release:main:minor": "node scripts/release_main minor",
+    "release:next": "node scripts/release_next",
```
sandrina-p, line 40: question: What's on your mind with this `next` release? 

Currently we have 2 different types of release:
- dev: to be run as many times as needed during development/PRs. eg `0.10.2-dev.20230619163824`
- main: to be run after a PR is merged and release a stable version `0.10.2-beta.0`

In your terminal you can run `git tags` to see all tags (releases) done so far. [or in github tags page](https://github.com/remoteoss/json-schema-form/tags).

Now you create `release:next`, what does it mean? I saw you already pushed [next-v1.0.0-alpha.1](https://github.com/remoteoss/json-schema-form/releases/tag/next-v1.0.0-alpha.1). Why the `next-` suffix? Why is it called `alpha` but not `beta`?

How is it different from the existing dev/main versions?

sandrina-p: question: How did you test this end-to-end? Did you use any other repo to test it works well (the full process including the actual release).

sandrina-p: I tried to run this locally and it went from `1.0.0-alpha.1` to `1.0.1`. Is this correct? Following the existing flow it should go to `dev` version/tag as it happens to v0.

image

But if the plan is to change the entire release process, then forgive me, I missed some thread/docs about it, as I don't understand this.

dragidavid: This issue is fixed with the latest changes 
Screenshot 2025-01-24 at 13 37 20

scripts/release_v1.js
```diff
@@ -0,0 +1,137 @@
...
+  const prefix = `v1-${releaseType}`;
+
+  let cmd;
+  if (releaseType === 'beta') {
+    // For beta, we commit package.json changes
+    cmd = `git add next/package.json && git commit -m "Release ${prefix} ${newVersion}" && git tag ${prefix}-${newVersion} && git push && git push origin --tags`;
+  } else {
+    // For dev, we only create a tag
+    cmd = `git tag ${prefix}-${newVersion} && git push origin --tags`;
+  }
```
dragidavid, line 85: Do not update the `package.json` if we're running the dev script

scripts/release_v1.js
```diff
@@ -0,0 +1,137 @@
...
+const packageJson = require(packageJsonPath);
+
+async function checkGitBranchAndStatus() {
+  console.log('Checking your branch...');
+  const resultBranch = await runExec('git branch --show-current', {
+    silent: true,
+  });
+  const branchName = resultBranch.stdout.toString().trim();
+  if (branchName === 'main') {
+    console.error(`🟠 You are at "main". Are you sure you wanna release a v1 version here?`);
```
sandrina-p, line 25: todo: @dragidavid this runs the exact same script for `dev` and for `beta`, but it cannot be the same.

- For `dev`, it cannot be run it on `main` branch.
- For `beta`, it must be run on `main` branch, and the local main must be up to date.

scripts/release_v1.js
```diff
@@ -0,0 +1,137 @@
...
+
+async function bumpVersion({ newVersion, releaseType }) {
+  // For beta, update package.json
+  if (releaseType === 'beta') {
+    const cmd = `cd next && npm version --no-git-tag-version ${newVersion}`;
+    await runExec(cmd);
+  }
+  // For dev, temporarily update version for publishing only
+  else {
+    const cmd = `cd next && npm --no-git-tag-version version ${newVersion} --no-workspaces-update --workspace-update=false`;
```
sandrina-p, line 63: q: What is this `--no-workspaces-update --workspace-update=false` for? Why do we need it for dev but not for main (beta)?

scripts/release_v1.js
```diff
@@ -0,0 +1,137 @@
...
+
+  const answer = await askForConfirmation('Ready to commit and publish it?');
+
+  if (answer === 'no') {
+    process.exit(1);
+  }
+
+  await checkNpmAuth();
+  const otp = await askForText('🔐 What is the NPM Auth OTP? (Check 1PW) ');
+
```
sandrina-p, line 130: todo: It's missing to update the CHANGELOG when it's a beta version (as we do in `release_main`)

scripts/release_v1.js
```diff
@@ -0,0 +1,137 @@
...
+    console.log('Creating new dev version...');
+    return `1.0.0-dev.${timestamp}`;
+  }
+
+  // For beta releases
+  console.log('Creating new beta version...');
+  if (currentVersion.includes('-beta.')) {
+    return semver.inc(currentVersion, 'prerelease', 'beta');
+  }
+  return '1.0.0-beta.0';
```
sandrina-p, line 52: bug: this return does not make sense. I'd throw an error instead.

next/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "1.0.0-alpha.1",
+  "version": "1.0.0-beta.2",
```
sandrina-p, line 3: todo: We also need a script to verify the package version of `next/` folder, similar to `pr_dev_version.js`. We do not want PRs being merged with `-dev` versions.

scripts/release_v1.js
dragidavid: Fixed in https://github.com/remoteoss/json-schema-form/pull/114/commits/4e653d1832c8b202d97ed9bfc4740bf4f2fce2a3

scripts/release_v1.js
dragidavid: Thats a good catch, it is not necessary 💡 The whole if statement is redundant for that matter.

Fixed in https://github.com/remoteoss/json-schema-form/pull/114/commits/a0bdf45f0aed30e7a218c9b7d16132f5a56f1a0e

scripts/release_v1.js
dragidavid: Added the changelog update based on the main script. This also adds `generate-changelog` package to `next`

Here https://github.com/remoteoss/json-schema-form/pull/114/commits/00e467a7fc38e6f38b08bfcd36d97c2d3f36b8d6

next/package.json
dragidavid: Added this!

Making it fail intentionally: https://github.com/remoteoss/json-schema-form/pull/114/commits/9259d8d63779f3e5ac3170217f167e5d86afa015

Fixing the intentional mistake above: https://github.com/remoteoss/json-schema-form/pull/114/commits/d11dc6793b74ab57d2c112d212a6ce3e7ec011a4

scripts/release_v1.js
sandrina-p: The file needs `next/CHANGELOG.md` needs to exist. It could be empty with just one line saying "Coming soon..." so something

sandrina-p: Nevermind, actually works well even if it doesn't exist. ✅

scripts/release_v1.js
sandrina-p: perfect!

next/package.json
sandrina-p: ahahah I see what you've done there 😅 it looks great!

sandrina-p, line 3: Can you change this to `-beta.0`? (this and the lock file)

And then we are ready to go! 🚢

dragidavid: I think the pnpm lock file didn't have the version so we should be good!

sandrina-p: even better. approved!

sandrina-p (approved):

PR #116, opened 2025-01-27
chore(next): define JsfSchema type extending JSONSchema

lukad:
This defines a `JsfSchema` type extending `JSONSchema`.

### DIFF
--- a/next/src/types.ts
+++ b/next/src/types.ts
@@ -1,0 +1,9 @@
+import type { JSONSchema } from 'json-schema-typed/draft-2020-12'
+
+export type JsfSchema = Exclude<JSONSchema, boolean> & {
+  'x-jsf-logic'?: {
+    validations: Record<string, object>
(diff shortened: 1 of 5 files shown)


next/src/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { JSONSchema } from 'json-schema-typed/draft-2020-12'
```
dragidavid, line 1: Assuming this is fine since its later than draft7

next/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "target": "ES2020",
     "module": "ESNext",
+    "moduleResolution": "bundler",
```
dragidavid, line 5: Thanks for adding this. Missed it out on my PRs

dragidavid (approved):

PR #118, opened 2025-02-03
feat(next): string validation

lukad:
This implements string validation (minLength, maxLength, pattern) as well as the groundwork for validating objects.
I will address building fields, as well as `formErrors` compatibility with v0 in a follow up PR.


next/test/validation/boolean_schema.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, expect, it } from '@jest/globals'
+import { createHeadlessForm } from '../../src'
+
+describe('boolean schema validation', () => {
+  it('returns an error if the value is false', () => {
+    const form = createHeadlessForm({ properties: { name: false } })
+
+    expect(form.handleValidation({ name: 'anything' })).toMatchObject({ formErrors: { '.name': 'always fails' } })
```
lukad, line 8: `formErrors: { '.name': 'always fails' }` does not match what v0 returns. I'll fix that in another commit.

next/src/validation/schema.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const typeValidationErrors = validateType(value, schema)
+  if (typeValidationErrors.length > 0) {
+    return typeValidationErrors
+  }
+
+  const errors = [
+    ...validateObject(value, schema),
+    ...validateString(value, schema),
+  ]
```
dragidavid, line 103: Assuming this is where we would list all our new validation functions. 

One thing I wanted to ask is would we keep using yup for validation or do we go vanilla for everything?

next/src/validation/string.ts
dragidavid, line 1: I like this approach where we have a file for each type. Easy to understand and probably to maintain in the future.

next/test/validation/string.test.ts
dragidavid, line 1: We can probably document that we'll have a test matching each validation file. Seems very straightforward but probably worth noting regardless.

next/src/types.ts
```diff
@@ -1,10 +1,37 @@
 import type { JSONSchema } from 'json-schema-typed/draft-2020-12'
 
+/**
+ * Defines the type of a `Field` in the form.
+ */
+export type JsfSchemaType = Exclude<JSONSchema, boolean>['type']
```
dragidavid, line 6: I know we went through this on the meet earlier today but can you remind me again when and how the type of JSONSchema is a boolean? 🤔

dragidavid (commented):
Left a few comments but largely I think this is a good approach. I guess we'll see how everything falls into place as we work on the upcoming parts.

Last question about testing:
I assume when it comes to testing, we should run the tests from v0 against the new code + run the new ones we write, right?

next/src/validation/schema.ts
lukad: > One thing I wanted to ask is would we keep using yup for validation or do we go vanilla for everything?
I think we should ditch yup, IMO it does not really bring much value.

lukad: @thien-remote 
> Is there a case where we would expect errors from multiple validation here?

Good point. This is just a stylistic choice. I think there two ways to do this:
1. Have one if - else if block in `validateSchema` that checks what type the schema specifies and call the corresponding validate function
2. Call all validate functions where they do that check and will either validate or not

I'm leaning towards the seconds because I'm thinking there might be cases where the condition whether to run the validations is more complex. Not sure though tbh, it's just a hunch. I don't feel strongly about this.

next/src/types.ts
lukad: Yeah, the naming here is tricky 😅. Indeed a JSON Schema can be `true` or `false`. That also applies to subschemas, i.e. object properties. `{ "properties": { "address": { "street": false } }, "required": ["street"]}` for example will always fail validation on `.address.street`.

`JsfSchemaType` is a type that the compiler should end up expanding to `"string" | "number" | "integer" | "array" | "object"`. These are the valid JSON Schema types. A boolean schema does not have a type, so that's why we have to exclude boolean.

next/src/validation/object.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param schema - The schema to validate against
+ * @returns An array of validation errors
+ * @description
+ * Validates each property of object against the schema while keeping track of the path to the property.
+ * Each property is validated with `validateSchema`.
+ */
+export function validateObject(value: SchemaValue, schema: NonBooleanJsfSchema): ValidationError[] {
+  if (typeof schema === 'object' && schema.properties && typeof value === 'object') {
+    const errors = []
+    for (const [key, propertySchema] of Object.entries(schema.properties)) {
```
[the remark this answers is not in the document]
lukad: I think that would be equivalent.

next/src/validation/schema.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @returns An array of validation errors
+ * @description
+ * - If the schema type is an array, the value must be an instance of one of the types in the array.
+ * - If the schema type is a string, the value must be of the same type.
+ */
+function validateType(value: SchemaValue, schema: JsfSchema): ValidationError[] {
+  const schemaType = getSchemaType(schema)
+  const valueType = value === undefined ? 'undefined' : typeof value
+
+  if (Array.isArray(schemaType) ? !schemaType.includes(valueType) : valueType !== schemaType) {
```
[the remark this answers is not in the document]
lukad: Yeah, agreed.

Is there a better way to write it than this?

```
  if (Array.isArray(schemaType) && !schemaType.includes(valueType)) {
    return [{ path: [], validation: 'type', message: `should be ${schemaType}` }]
  }

  if (valueType !== schemaType) {
    return [{ path: [], validation: 'type', message: `should be ${schemaType}` }]
  }
```

next/src/types.ts
dragidavid: Ah yes, thanks Thien! Somewhat agree on the naming but we can change that later once there is a better idea.

next/src/types.ts
```diff
@@ -1,10 +1,37 @@
 import type { JSONSchema } from 'json-schema-typed/draft-2020-12'
```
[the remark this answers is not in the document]
lukad: `@types/json-schema` is pretty outdated and only does JSON Schema Draft 04.

next/src/validation/string.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @description
+ * - Validates the string length against the `minLength` and `maxLength` properties.
+ * - Validates the string pattern against a regular expression defined in the `pattern` property.
+ */
+export function validateString(value: SchemaValue, schema: NonBooleanJsfSchema): ValidationError[] {
+  const errors: ValidationError[] = []
+
+  if (getSchemaType(schema) === 'string' && typeof value === 'string') {
+    if (schema.minLength && value.length < schema.minLength) {
+      errors.push({ path: [], validation: 'minLength', message: 'must be at least 3 characters' })
```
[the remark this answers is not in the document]
dragidavid: Ah, this is touching on my comment above https://github.com/remoteoss/json-schema-form/pull/118#discussion_r1939679850 about the current errors we get after validation.

next/src/types.ts
```diff
@@ -1,10 +1,37 @@
...
 
+/**
+ * Defines the type of a `Field` in the form.
+ */
+export type JsfSchemaType = Exclude<JSONSchema, boolean>['type']
+
+/**
+ * Defines the type of a value in the form that will be validated against the schema.
+ */
+export type SchemaValue = string | number | ObjectValue | undefined
```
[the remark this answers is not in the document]
lukad: No, this is unrelated to the schema-boolean confusion 😬
This is the type a concrete value can have that we validate against a schema. I think that number actually slipped in by accident. For now we only validate strings and objects. If it turns out that this is too strict we can drop it later.

next/src/form.ts
```diff
@@ -1,5 +1,98 @@
...
+  /**
+   * The message of the validation error
+   * @example
+   * 'is required'
+   */
+  message: string
+}
+
+export interface ValidationResult {
+  formErrors: Record<string, string> | undefined
```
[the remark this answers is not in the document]
lukad: For v0 compatibility it must be `undefined` 😅. At least I think it has to be `undefined` here because v0 ends up returning an object without the `formErrors` key when there are no errors.

next/src/form.ts
```diff
@@ -1,5 +1,98 @@
...
+
+  const handleValidation = (value: SchemaValue) => {
+    const result = validate(value, schema)
+    return result
+  }
+
+  return {
+    fields: [],
+    isError,
+    error: null,
```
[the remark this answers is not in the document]
lukad: No, it's just not implemented yet.

next/src/types.ts
```diff
@@ -1,10 +1,37 @@
...
 export type JsfSchema = Exclude<JSONSchema, boolean> & {
-  'properties'?: Record<string, JsfSchema> | boolean
+  'properties'?: Record<string, JsfSchema>
   'x-jsf-logic'?: {
     validations: Record<string, object>
     computedValues: Record<string, object>
   }
   'x-jsf-order'?: string[]
-}
+} | boolean
```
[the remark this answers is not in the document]
lukad: Ah... good question. I think that's a mistake. I'll remove the Exclude an `| boolean`.

next/src/validation/schema.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Get the type of a schema
+ * @param schema - The schema to get the type of
+ * @returns The type of the schema, or an array of types if the schema is an array. Will fallback to 'object' if no type is defined.
+ * @example
+ * getType(false) // 'boolean'
+ * getType(true) // 'boolean'
+ * getType({ type: 'string' }) // 'string'
+ * getType({ type: ['string', 'number'] }) // ['string', 'number']
```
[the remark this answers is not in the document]
lukad: When it's an array it means that the value can be either of the types.

next/src/validation/schema.ts
lukad: > Just one thing we need to keep in mind is that currently the yupSchema is responsible for the "nice errors" that we get after validation. We'll need to keep those somehow.

> I suppose the plan is to actually implement them? Sorry Luka but this I need you 😆

Ah sorry, forgot to hit send on a comment I had already written 😅
Yes, we need to return the same errors indeed. I'll follow up on this MR with building fields for the schema. That's where would also like to address this.

next/src/form.ts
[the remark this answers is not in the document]
lukad: Thanks for the insight Thien. I'll turn this into `?` then.

next/src/validation/object.ts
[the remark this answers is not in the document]
lukad: This is just me not being a TS dev :p

next/src/validation/string.ts
```diff
@@ -0,0 +1,53 @@
+import type { ValidationError } from '../form'
```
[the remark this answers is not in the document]
lukad: Yeah, you're probably right. I was thinking this would make it easy to extend but it just makes it hard to read.

dragidavid:
Approving this as it seems we all agree that this is a good starting point! Thanks for adding it Luka 🙌

dragidavid (approved):

next/src/validation/string.ts
lukad: Yup, I'll add this to the readme 👍

next/test/validation/string.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        '.name': 'should be string',
+      },
+    })
+  })
+
+  it('validates the string length against the minLength property', () => {
+    const result = createHeadlessForm({
+      type: 'object',
+      properties: {
+        name: { type: 'string', minLength: 3 },
```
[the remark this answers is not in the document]
lukad: I added a test case where `minLength` is `0`. Same for `maxLength`. And the `maxLength` validation does fail without also changing the `validateString` implementation to `if (schema.maxLength !== undefined && ...` from `if (schema.maxLength && ...`.

next/src/validation/string.ts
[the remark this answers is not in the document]
lukad: Now that I'm thought about it some more. How about we keep these specific validation error type definitions such as `StringValidationErrorType` where they are now but merge them only at the toplevel. I still feel like it's not a bad idea to define the errors where they are emitted.
Suggestion:

```
// form.ts
type ValidationErrorKind = ObjectValidationError | StringValidationError | NumberValidationError | etc...

type ValidationError {
  path: ...
  message: string,
  kind:
}
```

next/src/types.ts
lukad: > That being said, if we haven't supported it yet, it's wise to leave it out now.
We won't be able to escape having to deal with booleans on the types. That's just the way JSON schema is defined.

I think we have not supported boolean schemas before, but the actual implementation is just this:

```
  if (typeof schema === 'boolean') {
    return schema ? [] : [{ path: [], validation: 'valid', message: 'always fails' }]
  }
```

I'm fine with removing it though, but will have to check what v0 does in that case.

next/src/validation/schema.ts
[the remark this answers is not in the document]
dragidavid: We could potentially do something like this? `hasTypeMismatch` seems to be descriptive enough imo!
```
  const hasTypeMismatch = Array.isArray(schemaType)
    ? !schemaType.includes(valueType)
    : valueType !== schemaType;

  if (hasTypeMismatch) {
    return [{
      path: [],
      validation: 'type',
      message: `should be ${Array.isArray(schemaType) 
        ? schemaType.join(' | ') 
        : schemaType}`
    }];
  }
  ```

File: next/src/validation/schema.ts (hunk 2 above)
Comment (dragidavid) [reply next/src/validation/schema.ts:103]: Took me a good second to understand what you mean here Luka but I think I got it 😄 

>1. Have one if - else if block in validateSchema that checks what type the schema specifies and call the corresponding validate function
>2. Call all validate functions where they do that check and will either validate or not
>
>I'm leaning towards the seconds because I'm thinking there might be cases where the condition whether to run the validations is more complex. Not sure though tbh, it's just a hunch. I don't feel strongly about this.

I **think** also like the second approach here since it would keep this function clean (less if statements) and we could always just check in the individual validation function whether they need to run or not. Might be bit repetitive overall but it is still clearer overall IMO.

One thing that would change my opinion here is if that condition can change on the long term and become more complex in which case we would potentially need to update all the validation functions. You see where I'm going with this..

Comment (lukad) [reply next/src/validation/schema.ts:103]: I suggest we keep it like this then.

> One thing that would change my opinion here is if that condition can change on the long term and become more complex

I don't think this will be the case. But if we run into that problem, it won't be too bad. We won't have hundreds of validation functions.

File: next/src/form.ts [hunk 15]
```diff
@@ -1,5 +1,102 @@
...
+  const isError = validationResult !== null
+
+  const handleValidation = (value: SchemaValue) => {
+    const result = validate(value, schema)
+    return result
+  }
+
+  return {
+    fields: [],
+    isError,
```
lukad, line 98: According to the docs we only return `fields` and `handleValidation`. Do we need to keep isError and errors for backwards compatibility or can we drop them @sandrina-p?

next/src/types.ts
lukad: FWIW, I checked v0 using this schema `{ properties: { foo: false }, required: ['foo'] }`, and it just ignores boolean schemas. It returns no errors when validating `{ foo: "bar"}` but it also does not respect the required foo in this case. It will also return no errors for `{}`.
Let's just keep the boolean schema validation and fix that bug in the process.

lukad:
@dragidavid I just noticed I missed your question regarding tests.

> Last question about testing:
I assume when it comes to testing, we should run the tests from v0 against the new code + run the new ones we write, right?

The new tests should all pass. While we continue building `next` more and more v0 tests should pass as well.

next/src/validation/object.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Validate an object against a schema
+ * @param value - The value to validate
+ * @param schema - The schema to validate against
+ * @returns An array of validation errors
+ * @description
+ * Validates each property of object against the schema while keeping track of the path to the property.
+ * Each property is validated with `validateSchema`.
+ */
+export function validateObject(value: SchemaValue, schema: NonBooleanJsfSchema): ValidationError[] {
+  if (typeof schema === 'object' && schema.properties && typeof value === 'object') {
```
sandrina-p, line 18: @lukad / @dragidavid can we adopt the "early return" pattern anywhere we can? In this case it would be:

```js
if (typeof schema !== 'object' || !schema.properties || typeof value !== 'object') {
  return []
}

// the actual code...

```

next/src/form.ts
```diff
@@ -1,5 +1,102 @@
...
+  }, {})
+}
+
+interface CreateHeadlessFormOptions {
+  initialValues?: SchemaValue
+}
+
+export function createHeadlessForm(schema: JsfSchema, options: CreateHeadlessFormOptions = {}): FormResult {
+  const errors = validateSchema(options.initialValues, schema)
+  const validationResult = validationErrorsToFormErrors(errors)
```
sandrina-p, line 88: praise: Thanks for doing this, following the official spec of json schema errors, and turn them into "formErrors" ✨

next/src/form.ts
sandrina-p: We used them in the internal codebase, yes, please keep them!

next/src/validation/string.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @description
+ * - Validates the string length against the `minLength` and `maxLength` properties.
+ * - Validates the string pattern against a regular expression defined in the `pattern` property.
+ */
+export function validateString(value: SchemaValue, schema: NonBooleanJsfSchema): ValidationError[] {
+  const errors: ValidationError[] = []
+
+  if (getSchemaType(schema) === 'string' && typeof value === 'string') {
+    if (schema.minLength !== undefined && value.length < schema.minLength) {
+      errors.push({ path: [], validation: 'minLength', message: 'must be at least 3 characters' })
```
sandrina-p, line 33: 🐛 Question: Why is "3 characters hardcoded"? 🤔

next/src/validation/string.ts
```diff
@@ -0,0 +1,53 @@
...
+      errors.push({ path: [], validation: 'maxLength', message: `must be at most ${schema.maxLength} characters` })
+    }
+
+    if (schema.pattern !== undefined) {
+      const pattern = new RegExp(schema.pattern)
+      if (!pattern.test(value)) {
+        errors.push({
+          path: [],
+          validation: 'pattern',
+          message: `must match the pattern '${schema.pattern}'`,
```
sandrina-p, line 46: For UX reasons, we improve this message even more [v0 source](https://github.com/remoteoss/json-schema-form/blob/main/src/yupSchema.js#L394). `Must have a valid format. E.g. ${randomPlaceholder}`

next/test/validation/object.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, expect, it } from '@jest/globals'
+import { createHeadlessForm } from '../../src'
+
+describe('object schema validation', () => {
+  it('returns an error if the value is not an object', () => {
```
sandrina-p, line 5: nit: Could we keep the same naming convention as v0 tests. Instead of saying `if`, saying `when` or `given`

next/test/validation/object.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    const schema = { properties: { address: { type: 'object', properties: { street: { type: 'string' } } } } }
+    const form = createHeadlessForm(schema)
+
+    expect(form.handleValidation({})).not.toHaveProperty('formErrors')
+    expect(form.handleValidation({ address: {} })).not.toHaveProperty('formErrors')
+    expect(form.handleValidation({ address: 'not an object' })).toMatchObject({
+      formErrors: { '.address': 'should be object' },
```
sandrina-p, line 12: bug/question: Why does the path has a `.` prefix. This is different from v0 and would break the internal code.

next/test/validation/object.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    })
+  })
+
+  it('validates the object properties', () => {
+    const schema = { properties: { address: { type: 'object', properties: { street: { type: 'string' } } } } }
+    const form = createHeadlessForm(schema)
+
+    expect(form.handleValidation({ address: { street: 'some street' } })).not.toHaveProperty('formErrors')
+    expect(form.handleValidation({ address: { street: 10 } })).toMatchObject({
+      formErrors: { '.address.street': 'should be string' },
```
sandrina-p, line 22: bug/question: @lukad / @dragidavid  do you realize that these error messages are different from v0? This will fail all tests. The error messages MUST be the same, otherwise a lot of tests (here and in internal codebases) will fail.

If we want to change the errors message, let that for later as an opt-in config or something.

dragidavid: Yes, I mentioned this in another thread which I can't find now but after all we want all the error messages to be the same as before, definitely!

PR #119, opened 2025-02-05
DEVXP-2570: feat(next): Add custom order

thien-remote:
This PR adds support for custom order in the Next version. See:
- [Public docs](https://json-schema-form.vercel.app/?path=/docs/api-reference-json-schema--docs#x-jsf-order) (from Prev version)
- [Internal ticket](https://linear.app/remote/issue/DEVXP-2570/implement-x-jsf-order)

There are actually 2 parts actually, reflecting in the 2 commits:
- https://github.com/remoteoss/json-schema-form/commit/12336de1fb2b0fece1f4092f6090b89b87b557b7 adds the basic support for "fields" ([docs](https://json-schema-form.vercel.app/?path=/docs/api-reference-api--docs#field)), especially the recursive support for "object" type.
- https://github.com/remoteoss/json-schema-form/commit/018b0b8c619a6a47cb543c2e491cecf38ea04325 adds the actual custom order feature.

Do tell if you want me to split the PR as the commits 👋

### DIFF
--- a/next/src/custom/order.ts
+++ b/next/src/custom/order.ts
@@ -1,0 +1,53 @@
+import type { Field } from '../field/type'
+import type { JsfSchema } from '../types'
+
+function sort(params: {
+  fields: Field[]
+  order: string[]
+}): Field[] {
+  const { fields: prevFields, order } = params
+
+  // Map from field name to expected index
+  const indexMap: Record<string, number | undefined> = {}
+  order.forEach((key, index) => {
+    indexMap[key] = index
+  })
+
+  const nextFields = prevFields.sort((a, b) => {
+    // Compare by index
+    const indexA = indexMap[a.name] ?? Infinity
+    const indexB = indexMap[b.name] ?? Infinity
+
+    // The else actually only happens when both are Infinity,
+    // i.e., not specified in the order array
+    if (indexA !== indexB)
+      return indexA - indexB
+
+    // If not specified, maintain original relative order
+    return prevFields.indexOf(a) - prevFields.indexOf(b)
+  })
+
+  return nextFields
+}
+
+/**
+ * Sort fields by schema's `x-jsf-order`
+ */
+export function setCustomOrder(params: {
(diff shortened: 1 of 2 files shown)


lukad (approved):
🙌

next/src/custom/order.ts
```diff
@@ -0,0 +1,53 @@
...
+  fields: Field[]
+}): Field[] {
+  const { schema, fields } = params
+
+  // TypeScript does not yield if we remove this check,
+  // but it's only because our typing is likely not right.
+  // See internal discussion:
+  // - https://remote-com.slack.com/archives/C02HTN0LY02/p1738745237733389?thread_ts=1738741631.346809&cid=C02HTN0LY02
+  if (typeof schema === 'boolean')
+    throw new Error('Schema must be an object')
```
lukad, line 47: I don't think we should raise here, but `return fields` as they are instead.

lukad, line 47: I don't think we should raise here, but return fields as they are instead.

lukad: On the other hand schema can't ever be a boolean here, right? Because `setCustomOrder` is only called for object schemas in buildFieldObject. If `(typeof schema === 'boolean')` is true it means there is a bug in `buildFieldObject`. Maybe raising here is good thing then.

lukad: We could also specify the type of schema as `NonBooleanJsfSchema`. It should work without any other changes because `buildFieldObject` already checks whether we have a boolean schema.
We could also do the same in `buildFieldObject` & `buildSingle` respectively.

dragidavid: Agree with the last point, I ended up using the `NonBooleanJsfSchema` type as well after making sure its not a boolean.

dragidavid:
Happy to approve this since it seems that it covers the base cases. We can always adjust if needed.

dragidavid (approved):

PR #121, opened 2025-02-05
DEVXP-2603: chore(next): use json schema test suite

lukad:
This adds the json schema test suite repo as a submodule and defines a test for each test case from the test suite dynamically.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "next/json-schema-test-suite"]
+	path = next/json-schema-test-suite
+	url = git@github.com:json-schema-org/JSON-Schema-Test-Suite.git
--- a/next/test/validation/json_schema_test_suite.test.ts
+++ b/next/test/validation/json_schema_test_suite.test.ts
@@ -21,8 +21,8 @@ interface TestSchema {
 expect.extend({
   toBeValid(received: JsfSchema, value: SchemaValue, valid: boolean = true) {
     const form = createHeadlessForm(received, { initialValues: value })
-    const formErrors = form.handleValidation(value)
-    const pass = formErrors.formErrors !== undefined !== valid
+    const validationResult = form.handleValidation(value)
+    const pass = validationResult.formErrors !== undefined !== valid
     return {
       pass,
       message: () => `expected ${util.inspect(value)} ${valid ? 'to' : 'not to'} be valid for ${util.inspect(received)}`,
--- a/next/test/validation/json_schema_test_suite.test.ts
+++ b/next/test/validation/json_schema_test_suite.test.ts
@@ -1,4 +1,3 @@
-import type { ValidationResult } from '../../src/form'
 import type { JsfSchema, SchemaValue } from '../../src/types'
 import fs from 'node:fs'
 import path from 'node:path'
@@ -22,7 +21,8 @@ expect.extend({
   toBeValid(received: JsfSchema, value: SchemaValue, valid: boolean = true) {
     const form = createHeadlessForm(received, { initialValues: value })
     const validationResult = form.handleValidation(value)
-    const pass = validationResult.formErrors !== undefined !== valid
+    const hasFormErrors = validationResult.formErrors !== undefined
+    const pass = hasFormErrors !== valid
     return {
       pass,
       message: () => `expected ${util.inspect(value)} ${valid ? 'to' : 'not to'} be valid for ${util.inspect(received)}`,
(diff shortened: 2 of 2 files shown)


next/test/validation/json_schema_test_suite.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      pass,
+      message: () => `expected ${util.inspect(value)} ${valid ? 'to' : 'not to'} be valid for ${util.inspect(received)}`,
+    }
+  },
+})
+
+describe('JSON Schema Test Suite', () => {
+  const testsDir = path.join(__dirname, '..', '..', 'json-schema-test-suite', 'tests', 'draft2020-12')
+  const testFiles = fs.readdirSync(testsDir).filter(file => file.endsWith('.json'))
```
dragidavid, line 35: Thats a very clever solution 🙌

next/test/validation/json_schema_test_suite.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  description: string
+  schema: JsfSchema
+  tests: Test[]
+}
+
+expect.extend({
+  toBeValid(received: JsfSchema, value: SchemaValue, valid: boolean = true) {
+    const form = createHeadlessForm(received, { initialValues: value })
+    const formErrors = form.handleValidation(value)
+    const pass = formErrors.formErrors !== undefined !== valid
```
antoniocapelo, line 25:
```suggestion
    const validationResult = form.handleValidation(value)
    const pass = validationResult.formErrors !== undefined !== valid
```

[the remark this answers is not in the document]
lukad: I changed it to
```
    const hasFormErrors = validationResult.formErrors !== undefined
    const pass = hasFormErrors !== valid
```

next/test/validation/json_schema_test_suite.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  const testFiles = fs.readdirSync(testsDir).filter(file => file.endsWith('.json'))
+
+  for (const file of testFiles) {
+    const testFile: TestSchema[] = JSON.parse(fs.readFileSync(path.join(testsDir, file), 'utf8'))
+
+    for (const testSchema of testFile) {
+      describe(testSchema.description, () => {
+        for (const test of testSchema.tests) {
+          it(test.description, () => {
+            expect(testSchema.schema).toBeValid(test.data, test.valid)
```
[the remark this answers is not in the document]
lukad: Ok, I'll skip it with a TODO then. I tried to do that interface extension yesterday but couldn't figure out the correct types and cursor couldn't either.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "next/json-schema-test-suite"]
```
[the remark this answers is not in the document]
sandrina-p: TIL about gitmodules 💡

next/test/validation/json_schema_test_suite.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  valid: boolean
+}
+
+interface TestSchema {
+  description: string
+  schema: JsfSchema
+  tests: Test[]
+}
+
+expect.extend({
```
sandrina-p, line 20: q: How do I run these tests? Obviously it says the test-suit direction does not exist, what am I missing 🤔  (And add it to the README instructions)

next/test/validation/json_schema_test_suite.test.ts
sandrina-p: Indeed, TIL about this, super cool!

I couldn't run the tests though (my comment above), but I'm assuming most of them fail. We need a white-list of the ones we should skip, and gradually reduce the list as we go.

Another question: What if the source changes? We are pointing it to master, but shouldn't we point to a specific commit/version?

antoniocapelo (approved):
approved since the new suite is skipped and we can iterate in future PRs

next/test/validation/json_schema_test_suite.test.ts
lukad: Good point, because of the git submodule you will have to either
* run `git submodule update --init` in your existing repo
* or clone the repo with `git clone git@github.com:remoteoss/json-schema-form.git --recursive`
I will add that to the next Readme.

lukad: I'll do a follow up PR for that.

PR #127, opened 2025-02-07
DEVXP-2605: feat(next): implement x-jsf-presentation

lukad:
This implements basic features of `x-jsf-presentation`.
I've made some changes to existing code:
* renamed `buildFieldsSchema` to `buildFieldSchema`
  * it returns a single field
  * if the schema is an object schema it calls `buildFieldObject`
* the `buildFields` function now calls `buildFieldObject` and returns its `fields`


next/src/types.ts
```diff
@@ -36,3 +48,11 @@ export type JsfSchema = JSONSchema & {
...
  */
 export type NonBooleanJsfSchema = Exclude<JsfSchema, boolean>
+
+/**
+ * JSON Schema Form type specifically for object schemas.
+ * This type ensures the schema has type 'object'.
+ */
+export type JsfObjectSchema = NonBooleanJsfSchema & {
+  type: 'object'
+}
```
lukad, line 58: I've introduced this type so we can enforce at the type level schemas being passed to `createHeadlessForm` and certain other functions are object schemas.

next/test/form.test.ts
```diff
@@ -5,14 +5,4 @@ describe('createHeadlessForm', () => {
...
-
-  it('returns empty result given an empty schema', () => {
-    const result = createHeadlessForm({}, { initialValues: {} })
-
-    expect(result).toMatchObject({
-      fields: [],
-    })
-    expect(result.isError).toBe(false)
-    expect(result.error).toBeFalsy()
-  })
```
lukad, line 17: Now that empty schemas can't be passed to `createHeadlessForm` anymore I've removed this test. It was not that meaningful anyway.

next/test/validation/json-schema-test-suite.test.ts
```diff
@@ -20,10 +20,8 @@ interface TestSchema {
 
 expect.extend({
   toBeValid(received: JsfSchema, value: SchemaValue, valid: boolean = true) {
-    const form = createHeadlessForm(received, { initialValues: value })
-    const validationResult = form.handleValidation(value)
-    const hasFormErrors = validationResult.formErrors !== undefined
-    const pass = hasFormErrors !== valid
+    const errors = validateSchema(value, received)
```
lukad, line 23: I'm calling `validateSchema` directly here because `createHeadlessForm` now only takes `JsfObjectSchema`.

next/src/field/schema.ts
```diff
@@ -1,18 +1,58 @@
...
 
   if (schema.type === 'object') {
-    return buildFieldsObject({ schema })
+    const objectSchema: JsfObjectSchema = { ...schema, type: 'object' }
+    return buildFieldObject(objectSchema, name, required)
+  }
+
+  const inputType = getInputType(schema)
+
+  const field: Field = {
```
sandrina-p, line 41: bug: something is wrong in the lint, this should be a `let`, as we are reassigning it down below.

next/src/field/schema.ts
```diff
@@ -1,18 +1,58 @@
...
+    name: schema.title || name,
+    required,
+  }
+
+  if (schema.title !== undefined) {
+    field.label = schema.title
+  }
+
+  if (schema['x-jsf-presentation']?.accept) {
+    field.accept = schema['x-jsf-presentation']?.accept
```
sandrina-p, line 54: todo: we don't need this, as it comes already from line 22.

next/src/field/schema.ts
```diff
@@ -1,18 +1,58 @@
...
+    return buildFieldObject(objectSchema, name, required)
+  }
+
+  const inputType = getInputType(schema)
+
+  const field: Field = {
+    ...schema['x-jsf-presentation'],
+    inputType,
+    type: inputType,
+    name: schema.title || name,
```
sandrina-p, line 45: bug: is this correct? it must be the schema field key, but never its title 🤔

next/src/field/schema.ts
```diff
@@ -1,18 +1,58 @@
...
+    const objectSchema: JsfObjectSchema = { ...schema, type: 'object' }
+    return buildFieldObject(objectSchema, name, required)
+  }
+
+  const inputType = getInputType(schema)
+
+  const field: Field = {
+    ...schema['x-jsf-presentation'],
+    inputType,
+    type: inputType,
```
sandrina-p, line 44: todo: it's missing `jsonType: schema.type`

next/src/field/schema.ts
```diff
@@ -1,18 +1,58 @@
...
+    name: schema.title || name,
+    required,
```
sandrina-p, line 46: todo: We need to spread the reaming `...schema` down. This [internal ticket](https://linear.app/remote/issue/RMT-921) explains why

sandrina-p (changes requested):
Hi! This title caught my attention, so I came here with a quicker review ;)

next/src/field/schema.ts
lukad: Yeah, good catch this is indeed wrong. It should just be `name: name`. `schema.title` is used for `label`. I've fixed it.

next/src/field/schema.ts
lukad: 👍 just pushed another commit adding `jsonType`

next/src/field/schema.ts
lukad: Isn't assigning to keys of a `const` object allowed? We're not actually reassigning it.
Also my editor keeps changes it to `const` when it put `let` there 😅

next/src/field/schema.ts
```diff
@@ -1,18 +1,76 @@
...
+  }
+
+  if (Array.isArray(schema.type)) {
+    throw new TypeError('Array type is not yet supported')
+  }
+
+  const inputType = getInputType(schema)
+
+  const field: Field = {
+    ...schema['x-jsf-presentation'],
```
lukad, line 63: One thing I did here that I'm not enterely sure if it's correct is spread `x-jsf-presentation` to field before setting `inputType`, `type`, `name`, and so on. This means `x-jsf-presentation` attributes can't override those keys. This is however possible in v0.

next/src/field/schema.ts
lukad: Do we have a list of keys that should not be spread? Is it just `['title', 'oneOf']`?

next/src/field/schema.ts
```diff
@@ -1,18 +1,76 @@
...
+ * @param schema - The non boolean schema of the field
+ * @returns The JSON type for the field, based schema type. Default to 'text'
+ */
+function getJsonType(schema: NonBooleanJsfSchema): string {
+  if (Array.isArray(schema.type)) {
+    return 'select'
+  }
+
+  if (schema.type !== undefined) {
+    return schema.type as string
```
dragidavid, line 29: Do we need to cast this here? I'm wondering why it is not inferred as string by default 🤔

next/src/field/schema.ts
dragidavid: Yea I think its fine to use `const` here since we don't reassign the whole thing

next/test/validation/json-schema-test-suite.test.ts
dragidavid: This change wouldn't affect the other tests that use `createHeadlessForm` right? since those only take the schema as a parameter as well.

dragidavid:
Hey @lukad - the changes look good to me and makes sense. I'll let @sandrina-p go through this once more though since I might be missing something. This keyword is custom after all 😄

next/src/field/schema.ts
lukad: > I'm wondering why it is not inferred as string by default 🤔

Me too 😅
I think it's because `schema.type` is a `TypeValue`:

```
  export type TypeValue = (
    | ValueOf<TypeName>
    | TypeName
    | Array<ValueOf<TypeName> | TypeName>
    | ReadonlyArray<ValueOf<TypeName> | TypeName>
  );
```

we already know it's not an array, and I think it needs to be cast to a string because `TypeName` is an enum defined as

```
export enum TypeName {
  Array = "array",
  // etc.
}
```

next/test/validation/json-schema-test-suite.test.ts
lukad: Correct, atm the other tests all seem to pass an object schema to `createHeadlessForm` so they don't break.

next/src/field/schema.ts
dragidavid: Ah yea I think its because that type can be anything of that list so you need to be explicit. I'm sure there is a way around that so we can get rid of the `as string` part but its fine

dragidavid (approved):
Approving this for now. We can make any necessary changes later as I think those would be smaller tweaks anyway.

next/src/field/schema.ts
sandrina-p: In my opinion, we should spread everything, even `title` and `oneOf`. Minimize as few opinions as possible.

sandrina-p: P.S. I know, having field.title and field.label doesn't sound right, so let's leave title as the only one out for now. in the future i'd see this being customizable.

```js
createHeadlessForm(jsonSchema, {
  mapKeys: { title: 'label' }
}
```

So the Field would be indeed `label` but by default it should be `title` (without any opinionated renames)

next/src/field/schema.ts
sandrina-p: I agree with it. I don't remember any case where we would need to override those. And if we really need, then we'll catch it and fix it.

sandrina-p (approved):

next/src/field/schema.ts
lukad: Sounds good, I've added it now as the first thing assigned to a field so that the spread `x-jsf-presentation` and the explicitly mapped keys like `inputType`, `name`, etc. take precedence.

``` typescript
  const { title: _title, ...spreadSchema } = schema

  const field: Field = {
    ...spreadSchema,
    ...schema['x-jsf-presentation'],
    inputType,
    type: inputType,
    jsonType: getJsonType(schema),
    name,
    required,
  }
```

sandrina-p: It's great, thanks!

sandrina-p (approved):

PR #129, opened 2025-02-12
DEVXP-XXX: handle paths

dragidavid:
This PR applies a couple suggestions from [this previous PR](https://github.com/remoteoss/json-schema-form/pull/128), thanks @thien-remote.

Changes are:
- adds a new package `validator` to make the validations better in `format.ts`
- uses the `Format` enum from the `json-schema-typed` package instead of our custom solution (again in `format.ts)
- adjusts the way we handle the `path` key across the package - now matching the way we do it in the original package
- some minor changes, adding comments and tests

### DIFF
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -17,8 +17,13 @@ interface FormResult {
 export interface ValidationError {
   /**
    * The path to the field that has the error
+   * - For field-level errors: array of field names (e.g., ['address', 'street'])
+   * - For schema-level errors: empty array []
+   * - For nested validations: full path to the field (e.g., ['address', 'street', 'number'])
    * @example
-   * ['address', 'street']
+   * [] // schema-level error
+   * ['username'] // field-level error
+   * ['address', 'street'] // nested field error
    */
   path: string[]
   /**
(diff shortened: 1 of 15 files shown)


next/src/validation/schema.ts
```diff
@@ -8,30 +8,33 @@ import { validateString } from './string'
...
-   * The value fails validation against anyOf subschemas
+   * Composition keywords
    */
   | 'anyOf'
+  | 'oneOf'
+  | 'allOf'
+  | 'not'
+  | 'if'
+  | 'then'
+  | 'else'
```
dragidavid, line 26: added these in here since I'll be working on them (at least some of them) in the next few days!

next/src/validation/format.ts
```diff
@@ -4,51 +4,81 @@ import type { SchemaValidationErrorType } from './schema'
...
+  regex: (value: string) => {
+    try {
+      const pattern = new RegExp(value)
+      return pattern instanceof RegExp
+    }
+    catch {
+      return false
+    }
+  },
 }
```
dragidavid, line 71: I guess here we would benefit a lot from `yup` or `zod` 🤔

next/src/validation/schema.ts
```diff
@@ -1,37 +1,39 @@
 import type { ValidationError } from '../form'
 import type { JsfSchema, JsfSchemaType, SchemaValue } from '../types'
-import type { ObjectValidationErrorType } from './object'
```
dragidavid, line 3: Removed this since it is redundant. We were just doing:
`export type ObjectValidationErrorType = StringValidationErrorType`

next/src/validation/format.ts
dragidavid: Ended up changing this so it now uses `validator.js`

next/src/form.ts
```diff
@@ -17,8 +17,13 @@ interface FormResult {
 export interface ValidationError {
   /**
    * The path to the field that has the error
+   * - For field-level errors: array of field names (e.g., ['address', 'street'])
```
lukad, line 20: Adding/improving all this documentation is great 👍

lukad (approved):

PR #130, opened 2025-02-14
DEVXP-2537: validate numeric instances

lukad:
This adds validation of numbers and integers. Unfortunately this requires some special handling when validating types as `{'type': 'integer'}` expects a value which is of type `number` for which `Number.isInteger` is true.

I also noticed and fixed two subtle bugs:
1. integer/number or string validations must also be applied when their respective keys like `maxLength`, `minimum`, etc. are present when the schema specifies no type but the value is number/integer or string respectively
2. The custom `toBeValid` matcher was just checking if a validation had errors or not and was not taking into account whether it should have errors or not


dragidavid:
Hey @lukad - just had a look and the code itself make sense. Seems like we have the new tests passing as well but the number of failing `json-schema-test-suite` tests has also increased to 424 from 200 something. Is that expected here? 🤔 

Wondering if thats because of the bugs you found.

lukad:
> Hey @lukad - just had a look and the code itself make sense. Seems like we have the new tests passing as well but the number of failing `json-schema-test-suite` tests has also increased to 424 from 200 something. Is that expected here? 🤔
> 
> Wondering if thats because of the bugs you found.

Yup, I've just had another look at the currently failing tests and they seem to be failing because of things not implemented yet.
* additionalProperties
* allOf
* $anchor
* $ref & $dynamicRef
* const
* dependentSchemas
* enum
* all items / array related things
* not
* if/then/else

We won't even have to support the majority of these keywords/features, so I think it's looking pretty good.
There are some tests for edge cases for number and string validation which are failing. We can address those separately.

Some examples are `multipleOf` by small numbers

```
    by small number
      ✕ 0.0075 is multiple of 0.0001
      ✓ 0.00751 is not multiple of 0.0001
```

or string maxLength validation

```
    maxLength validation
      ✓ shorter is valid
      ✓ exact length is valid
      ✓ too long is invalid
      ✓ ignores non-strings
      ✕ two graphemes is long enough
```


> Side note: Wondering if we can adjust the json that we generate with the failed tests so its broken down by the thing that it is testing. So like we would have keys with `number` - `string` - `anyOf` etc.. and their value would be the array of failing tests in that "category". Would be easier to see whats up.

Yeah, that would be neat. It's a bit tedious to go through the failing tests, I'm not sure how we could organize it though 🤔

dragidavid:
> @dragidavid
> 
> > Hey @lukad - just had a look and the code itself make sense. Seems like we have the new tests passing as well but the number of failing `json-schema-test-suite` tests has also increased to 424 from 200 something. Is that expected here? 🤔
> > Wondering if thats because of the bugs you found.
> 
> Yup, I've just had another look at the currently failing tests and they seem to be failing because of things not implemented yet.
> 
> * additionalProperties
> * allOf
> * $anchor
> * $ref & $dynamicRef
> * const
> * dependentSchemas
> * enum
> * all items / array related things
> * not
> * patternProperties
> * if/then/else
> 
> We won't even have to support the majority of these keywords/features, so I think it's looking pretty good. There are some tests for edge cases for number and string validation which are failing. We can address those separately.
> 
> Some examples are `multipleOf` by small numbers
> 
> ```
>     by small number
>       ✕ 0.0075 is multiple of 0.0001
>       ✓ 0.00751 is not multiple of 0.0001
> ```
> 
> or string maxLength validation
> 
> ```
>     maxLength validation
>       ✓ shorter is valid
>       ✓ exact length is valid
>       ✓ too long is invalid
>       ✓ ignores non-strings
>       ✕ two graphemes is long enough
> ```
> 
> > Side note: Wondering if we can adjust the json that we generate with the failed tests so its broken down by the thing that it is testing. So like we would have keys with `number` - `string` - `anyOf` etc.. and their value would be the array of failing tests in that "category". Would be easier to see whats up.
> 
> Yeah, that would be neat. It's a bit tedious to go through the failing tests, I'm not sure how we could organize it though 🤔

I see, yep all good in that case, lets go 🚀 

I'll have a look at that json structure over the weekend. Still trying to wrap up the keyword issues.

dragidavid (approved):

PR #132, opened 2025-02-14
fix(next): fix min and max length for non ascii strings

lukad:
This uses the [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) to correctly calculate the string length for strings including multi byte characters.

CleanShot 2025-02-14 at 17 41 57@2x


next/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
```
lukad, line 3: Had to bump this in order to use `Intl.Segmenter`, is this right?

dragidavid:
Just checked this, looks great. Widely supported as well 🆗

dragidavid (approved):

PR #133, opened 2025-02-16
DEVXP-XXX: add missing keyword validation

dragidavid:
This PR adds the validation and tests for the 3 missing keywords, `allOf`, `oneOf` and `not`. 

The reason why I put all these together because as I was working on `allOf` separately I noticed that it mostly matches `anyOf` so I thought it would probably match `oneOf`. Made sense to put all of them together.

Also added a few tests of our own to test some extra cases on top of all the cases defined in the `json-schema-test-suite`. 

Happy to say that all of them are passing. ✅


lukad (approved):
Great stuff! And thanks for adding that `null` handling as well.

PR #137, opened 2025-02-18
feat(next): DEVXP-2564: implement conditional validation

lukad:
This PR
* implements `if`/`then`/`else`
* implements `const`
* and simplifies composition (`allOf`, `anyOf`, `oneOf`, `not`) a bit


next/src/validation/composition.ts
dragidavid, line 1: Hah! I would've never made this change simply because I didn't notice it at the time.

Thanks Luka

next/src/validation/conditions.ts
dragidavid, line 1: nit: Nothing to add about the code itself, just comments 😄 We should maybe follow the same approach as other components?

next/src/validation/const.ts
```diff
@@ -0,0 +1,56 @@
+import type { ValidationError } from '../form'
+import type { NonBooleanJsfSchema, SchemaValue } from '../types'
+
+function deepEqual(a: any, b: any): boolean {
```
dragidavid, line 4: nit: `any` 👀

next/test/validation/composition.test.ts
dragidavid, line 1: 🙌 Lot better, thank you!

next/src/validation/schema.ts
dragidavid, line 1: Again, great change! Thanks for simplifying. I went a little overboard it seems.

dragidavid (commented):
Couple nit comments but nothing is blocking. Maybe just add some comments?

lukad:
> Couple nit comments but nothing is blocking. Maybe just add some comments?

Yeah good point! I forgot to add comments.

next/src/validation/conditions.ts
lukad: I added comments to the functions I added. Totally forgot about that 😅

dragidavid (approved):

dragidavid (approved):

PR #138, opened 2025-02-18
feat(next): DEVXP-2564: implement enum validation

lukad:
This adds validation against enums. It is based on on #137, so I'm leaving it on draft until that is merged but it can already be reviewed.


next/src/validation/util.ts
```diff
@@ -0,0 +1,49 @@
...
+    return true
+  }
+
+  // If both are null, the check above has returned true.
+  // If one is null, we return false because we know they are not equal
+  // and since `typeof null === 'object'`, we must not let the null value
+  // pass through as an object.
+  if (a === null || b === null) {
+    return false
+  }
```
lukad, line 16: This is the only change I made to this function while moving it to a dedicated module. I didn't run into the edge case handled here with the json schema test suite tests for `const` but luckily with the `enum` tests.

dragidavid:
Seems good! I'll have a closer look tomorrow morning.

dragidavid (dismissed):
Seems good! Lets go 🚀

dragidavid (approved):

PR #141, opened 2025-02-20
DEVXP-2591: feat(next): standardize the results

dragidavid:
This PR attempts to migrate one of our existing forms to use the `next` version of `createHeadlessForm` and the validations.

At the moment we have the generated form/field structure matching the original version but the validation is not working yet. 

TODO:
- [x] fix field validation in `next`
- [x] remove console logs and clean up types


next/src/field/schema.ts
```diff
@@ -53,27 +101,50 @@ export function buildFieldSchema(
     return buildFieldObject(objectSchema, name, required)
   }
 
-  if (Array.isArray(schema.type)) {
+  if (schema.type === 'array') {
```
dragidavid, line 104: This is a fix that needs to be here. We got this wrong initially.

next/src/field/schema.ts
dragidavid, line 1: In general I'll need some help with this file, so it generates the same shape now as the main jsf version in prod but it obviously has its flaws. 

I mainly need some help understanding `computedAttributes`, `errorMessage` and the special fields like `maxLength`, `minDate` etc.. Which ones should be explicitly include in this file and the `Field` type? I want to make sure that I only include the things that we 100% need in order to make this next version work the same way.

next/test/fields.test.ts
dragidavid, line 1: Again, this is just the fix for this broken check.

next/src/field/object.ts
```diff
@@ -23,6 +23,8 @@ export function buildFieldObject(schema: JsfObjectSchema, name: string, required
 
   const orderedFields = setCustomOrder({ fields, schema })
 
+  // TODO: fix missing fields here
+  // QUESTION: why are we defaulting to a fieldset?
```
sandrina-p, line 27: For @lukad to answer. The default should be text (at least in v0 it's the default)

next/src/field/schema.ts
sandrina-p: - `errorMessage` is included when the json field has `x-jsf-errorMessage`.
- `maxLength` is when the json field has `maxLength`, same for similar ones.
- `computedAttributes` is when the json field has `json-logic`, not needed now.

next/src/field/schema.ts
```diff
@@ -32,12 +32,60 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
...
+/**
+ * TODO: go over the added code here and make sure it's correct and includes all the necessary fields
+ * this is a rough draft and needs to be cleaned up
+ */
+
+/**
+ * Get field options from schema
+ */
+function getFieldOptions(schema: NonBooleanJsfSchema, presentation: any) {
+  // Handle deprecated presentation.options first
```
sandrina-p, line 64: todo: No, no, let's **not support deprecated** features. If we find a JSON Schema using deprecated features, we fix them first.

next/src/field/schema.ts
```diff
@@ -32,12 +32,60 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
...
+  if (presentation.options) {
+    return presentation.options
+  }
+
+  // Handle oneOf or radio input type
+  if (schema.oneOf || presentation.inputType === 'radio') {
```
sandrina-p, line 70: Based on the comment above https://github.com/remoteoss/json-schema-form/pull/141/files#r1968082432, it means this does not need to look into the `inputType`

next/src/field/schema.ts
```diff
@@ -53,27 +101,50 @@ export function buildFieldSchema(
...
+    field.description = schema.description
+  }
+
+  if (schema.maxLength) {
+    field.maxLength = schema.maxLength
+  }
+
+  if (schema.format) {
+    field.format = schema.format
+  }
```
sandrina-p, line 141: todo: These if's are not scalable, we'll endup with a LOT of them. We should spread all schema attributes into field by default [as explained here](https://linear.app/remote/issue/RMT-921)

next/test/fields.test.ts
```diff
@@ -24,11 +24,11 @@ describe('fields', () => {
     ])
   })
 
-  it('should throw an error if the type is an array', () => {
+  it('should throw an error if the type equals "array"', () => {
```
sandrina-p, line 27: todo: we can be more explicit on what it means:

```suggestion
  it('should throw an error if the type equals "array" (group-array)', () => {
```

next/src/field/object.ts
dragidavid: Changed this to text for now.

next/src/field/schema.ts
dragidavid: Noted, this first schema is fine in that sense

next/src/field/schema.ts
```diff
@@ -32,12 +19,51 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
...
+    })
+}
+
+/**
+ * Get field options from schema
+ */
+function getFieldOptions(schema: NonBooleanJsfSchema) {
+  // Handle oneOf or radio input type
+  if (schema.oneOf) {
+    return convertToOptions(schema.oneOf || [])
```
antoniocapelo, line 49: We don't need this `|| []` fallback, right?

next/src/field/schema.ts
antoniocapelo: Agreed 👍

next/src/field/type.ts
```diff
@@ -1,19 +1,25 @@
...
   fields?: Field[]
   type: string
   inputType: string
   required: boolean
   jsonType: string
+  isVisible: boolean
   accept?: string
-  description?: string
-  maxFileSize?: number
+  errorMessage: Record<string, string>
```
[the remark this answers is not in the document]
antoniocapelo: (doing so, would reduce the number of tests changed, because these values can just be absent)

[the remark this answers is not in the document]
dragidavid: They are not optional in v0 and I'm not sure how other forms will "react" to these missing properties. We can try though as it won't affect the 1st form that we're changing.

next/test/fields.test.ts
```diff
@@ -107,19 +112,15 @@ describe('fields', () => {
...
-        'name': 'file',
-        'label': 'Some field',
-        'required': false,
-        'accept': '.pdf,.doc',
-        'maxFileSize': 5000000,
-        'x-jsf-presentation': {
-          inputType: 'file',
-          accept: '.pdf,.doc',
-          maxFileSize: 5000000,
-        },
```
dragidavid, line 122: We shouldn't have this in the final `field`, thats why I removed it.

next/src/field/type.ts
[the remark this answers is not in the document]
antoniocapelo: I'll let for @sandrina-p to reply, but it seems that, if the `x-jsf-error-message` and `x-json-logic` are optional, so should be the fields that originate from it, right?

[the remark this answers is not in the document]
sandrina-p: Exactly yes. The errorMesssge comes from the json schema directly. The computedAttributes is related to json-logic internals, not even needed for the final output (it was a mistake from v0)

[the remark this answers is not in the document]
dragidavid: Thanks team, I'll address that today 👌

next/src/field/object.ts
lukad: Text makes sense 👍

dragidavid:
Alright team, this is ready for another review. @lukad @sandrina-p

next/src/form.ts
```diff
@@ -114,6 +122,70 @@ function validationErrorsToFormErrors(errors: ValidationError[]): Record<string,
   }, {})
 }
 
+/**
+ * Apply custom error messages from the schema to validation errors
+ * @param errors - The validation errors
+ * @param schema - The schema that contains custom error messages
+ * @returns The validation errors with custom error messages applied
+ */
+function applyCustomErrorMessages(errors: ValidationError[], schema: JsfSchema): ValidationError[] {
```
dragidavid, line 131: My attempt to basically prioritize the `x-jsf-errorMessage` over the validation errors if exists. You might think its too much so please let me know if thats the case.

Works though. 😅

next/src/validation/schema.ts
```diff
@@ -93,19 +97,24 @@ function validateType(value: SchemaValue, schema: JsfSchema, path: string[] = []
...
-          validation: 'type',
-          message: `should be ${schemaType}`,
-        }]
+      : [
+          {
+            path,
+            validation: 'required',
+            message: 'is required',
+          },
+        ]
```
dragidavid, line 117: Apparently type validation should be prioritized over required validation but without this we would just end up always showing `should be string` instead of `is required` for fields that are required but have value of `null`.

A bit confusing so we'll probably revisit this when we implement readable/pretty error messages.

next/test/validation/required_field.test.ts
dragidavid, line 1: Added a couple of tests specifically to test for custom error messages and required fields.

lukad (approved):
🙌

Issue #115, opened 2025-01-23
select inputType and integer type


js-ta:
Hello,

in #50 I saw a similar problem to the one I have.
If I try to use the same schema, but with "type": "integer" instead of "type":"number", I get the error following error:

    Error: yupSchema.integer is not a function

This is reproducable in your playground.

Here is the schema:

    {
        "type": "object",
        "additionalProperties": false,
        "properties": {
            "AccountType": {
                "title": "Select account type",
                "description": "",
                "x-jsf-presentation": {
                    "inputType": "select"
                },
                "oneOf": [
                    {
                        "title": "Normal User",
                        "const": 0
                    },
                    {
                        "title": "Business",
                        "const": 1
                    },
                    {
                        "title": "Services Provider",
                        "const": 2
                    }
                ],
                "type": "integer"
            }
        },
        "required": [],
        "x-jsf-order": [
            "AccountType"
        ]
    }

js-ta:
Tried to quickly find what causes the problem:

`generateOptionSchema` does not handle an integer type. It is handled as string by default. Also `withInteger` is added to the validators which finally leads to the error.

When I extend `generateOptionSchema` by adding

    case 'integer':
        return yupSchemas.radioOrSelectNumber(optionValues);
        
and prevent `validators.push(withInteger)` from being called when the field is a select by adding

    const hasOptions = field.options?.length > 0;
    if (jsonType === 'integer' && !hasOptions) {
        validators.push(withInteger);
    }
    
I was able to workaround this issue.

sandrina-p:
Hi @js-ta 👋  Thank you a lot for the report and digging into the root cause. Could you open a PR with the fix and a unit test? 🙇

PR #142, opened 2025-03-04
fix(select): Support integer values

js-ta:
This PR is an attempt to fix the issue described in #115 to be able to use integer values in select elements.

It simply handles an integers like a number is handled already in a select and prevents the withInteger schema from being added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.11-beta.0",
+  "version": "0.11.12-dev.20250313084704",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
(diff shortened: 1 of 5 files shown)


sandrina-p:
Hi @js-ta, the team will review this some time next week, is that okay? :)

js-ta:
That's fine for me.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.11-beta.0",
+  "version": "0.11.12-dev.20250313084704",
```
sandrina-p, line 3: Hi @js-ta! We are building the next [major version of JSF](https://github.com/remoteoss/json-schema-form/tree/main/next), which will include your bugfix/feature. For that reason we won't review your PR, so we can focus on shipping the next version as soon as possible.

In the meantime, here's a dev release based on this PR to unblock you:
```
npm i @remoteoss/json-schema-form@0.11.12-dev.20250313084704
```
We hope that's okay for now. Thank you!

PR #146, opened 2025-03-10
chore(next): DEVXP-2630: run v0 tests

lukad:
This adds three scripts to run v0 tests and lets us keep track of which tests are failing/passing.

1. `test:v0-update-baseline` saves current v0 failing and passing tests to a json file that should be committed
2. `test:v0` runs v0 tests and saves the failing/passing tests to another file and runs `test:v0-compare-results`
3. `test:v0-compare-results` compares the the test output against the baseline

Output when tests previously passing are now failing:
CleanShot 2025-03-10 at 11 16 38@2x

Output when failing tests have been fixed:
CleanShot 2025-03-10 at 11 17 12@2x


next/test/v0_compare_test_results.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable no-console */
+// const fs = require('node:fs')
+// const process = require('node:process')
```
dragidavid, line 3: I think once we clean this file out and remove the commented lines, we're good to go

lukad: Done 👍

PR #148, opened 2025-03-11
chore(next): DEVXP-2682: centralize error messages

lukad:
This PR centralizes generating error messages.

1. `message` is removed from the `ValidationError` interface.
1. There is a new `addErrorMessages` function that 
1. `path` in the `ValidationError` interface now includes the full path to a subschema when a composition keyword like e.g. `anyOf` is used.
   This is needed to look up a schema & value pair in `addErrorMessages` given a path.

### DIFF
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -129,15 +129,27 @@ function getSchemaAndValueAtPath(rootSchema: JsfSchema, rootValue: SchemaValue, 
           currentValue = currentValue[Number(segment)]
         }
       }
+      // Skip the 'allOf', 'anyOf', and 'oneOf' segments, the next segment will be the index
       else if (segment === 'allOf' && currentSchema.allOf) {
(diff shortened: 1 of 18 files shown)


next/src/form.ts
```diff
@@ -100,19 +101,81 @@ function validationErrorsToFormErrors(errors: ValidationError[]): FormErrors | n
...
+      // Skip the 'allOf', 'anyOf', and 'oneOf' segments, the next segment will be the index
+      else if (segment === 'allOf' && currentSchema.allOf) {
+        continue
+      }
+      else if (segment === 'anyOf' && currentSchema.anyOf) {
+        continue
+      }
+      else if (segment === 'oneOf' && currentSchema.oneOf) {
+        continue
+      }
```
dragidavid, line 146: Good approach 👌

dragidavid (approved):
Really like this approach. Easier to understand as everything is all together.

PR #150, opened 2025-03-20
fix(next): RMT-1405: fix error messages for conditional schemas

lukad:
This addresses a bug where validation errors in conditional schemas would not generate correct error messages.
Check out the added tests for concrete examples.

### DIFF
--- a/next/src/validation/schema.ts
+++ b/next/src/validation/schema.ts
@@ -168,7 +168,6 @@ export function validateSchema(
     ...validateObject(value, schema, options, path),
     ...validateString(value, schema, path),
     ...validateNumber(value, schema, path),
-    ...validateCondition(value, schema, options, required, path),
     ...validateNot(value, schema, options, path),
     ...validateAllOf(value, schema, options, path),
     ...validateAnyOf(value, schema, options, path),
--- a/next/test/json-schema-test-suite/failed-json-schema-test-suite.json
+++ b/next/test/json-schema-test-suite/failed-json-schema-test-suite.json
@@ -1,37 +1,21 @@
 {
   "JSON Schema Test Suite": {
     "additionalProperties being false does not allow other properties": [
-      "no additional properties is valid",
-      "an additional property is invalid",
-      "ignores arrays",
-      "ignores strings",
-      "ignores other non-objects",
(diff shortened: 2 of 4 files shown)


next/test/json-schema-test-suite/failed-json-schema-test-suite.json
```diff
@@ -1,1770 +1,613 @@
 {
```
lukad, line 1: I think we accidentally committed this file with almost all tests failing.

next/src/validation/schema.ts
```diff
@@ -168,7 +168,6 @@ export function validateSchema(
     ...validateObject(value, schema, options, path),
     ...validateString(value, schema, path),
     ...validateNumber(value, schema, path),
-    ...validateCondition(value, schema, options, required, path),
```
lukad, line 171: We were calling `validateCondition` twice.

dragidavid: good catch

next/test/json-schema-test-suite/failed-json-schema-test-suite.json
dragidavid: Yea this is too big of a diff to ignore 😅 I think I found the culprit though. Seems that we "added" loads of failing tests in [this PR](https://github.com/remoteoss/json-schema-form/pull/148/files#diff-34797dfc0f7e77b6a74ea3518f9cc014e9c73a30d995d7854e4cd308f85a805c)

So I think this is _fine_

dragidavid (approved):
Did a bit of digging about the big diff in the json file and I think it is all good here. 

Looks great!

next/src/validation/schema.ts
lukad: While debugging the error message issue I was *very* confused why condition errors were always emitted twice. Didn't actually break anything though.

next/test/json-schema-test-suite/failed-json-schema-test-suite.json
lukad: we probably missed it because this stupid "Large diffs are not rendered by default." message is kinda easy to miss... 🙃

PR #151, opened 2025-03-25
feat(next): DEVXP-2565: implement visibility calculation

lukad:
The main two things this PR adds is:
* Setting `isVisible` based on conditional schema
* Setting `const` and `checkboxValue` on fields
* Simplifies `required` validation

### DIFF
--- a/next/src/field/object.ts
+++ b/next/src/field/object.ts
@@ -28,7 +28,7 @@ export function buildFieldObject(schema: JsfObjectSchema, name: string, required
     type: schema['x-jsf-presentation']?.inputType || 'fieldset',
     inputType: schema['x-jsf-presentation']?.inputType || 'fieldset',
     jsonType: 'object',
-    name: schema.title || name,
+    name,
     required,
     fields: orderedFields,
     isVisible: true,
--- a/next/src/field/schema.ts
+++ b/next/src/field/schema.ts
@@ -137,6 +137,14 @@ export function buildFieldSchema(
     ...(errorMessage && { errorMessage }),
   }
 
+  if (schema.const) {
+    field.const = schema.const
+
+    if (inputType === 'checkbox') {
+      field.checkboxValue = schema.const
+    }
+  }
+
   if (schema.title) {
     field.label = schema.title
   }
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -298,7 +298,6 @@ export function createHeadlessForm(
 
   const handleValidation = (value: SchemaValue) => {
     const result = validate(value, schema, options.validationOptions)
-    console.warn({ value, result })
     updateFieldVisibility(fields, value, schema, options.validationOptions)
     return result
   }
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -292,8 +292,9 @@ export function createHeadlessForm(
   schema: JsfObjectSchema,
   options: CreateHeadlessFormOptions = {},
 ): FormResult {
+  const initialValues = options.initialValues || {}
   const fields = buildFields({ schema })
-  updateFieldVisibility(fields, options.initialValues, schema)
+  updateFieldVisibility(fields, initialValues, schema)
   const isError = false
 
   const handleValidation = (value: SchemaValue) => {
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -1,10 +1,14 @@
-import type { ValidationError } from './errors'
+import type { ValidationError, ValidationErrorPath } from './errors'
 import type { Field } from './field/type'
 import type { JsfObjectSchema, JsfSchema, NonBooleanJsfSchema, SchemaValue } from './types'
+import type { ValidationOptions } from './validation/schema'
 import { getErrorMessage } from './errors/messages'
 import { buildFieldObject } from './field/object'
 import { validateSchema } from './validation/schema'
 import { isObjectValue } from './validation/util'
+import { updateFieldVisibility } from './visibility'
+
+export { ValidationOptions } from './validation/schema'
 
 interface FormResult {
   fields: Field[]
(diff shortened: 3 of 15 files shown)


next/src/field/object.ts
```diff
@@ -28,7 +28,7 @@ export function buildFieldObject(schema: JsfObjectSchema, name: string, required
     type: schema['x-jsf-presentation']?.inputType || 'fieldset',
     inputType: schema['x-jsf-presentation']?.inputType || 'fieldset',
     jsonType: 'object',
-    name: schema.title || name,
```
lukad, line 31: This just turned out to be wrong. `title` is the human readable name and `name` is the field name that will be used as the html input name.

next/src/form.ts
```diff
@@ -272,42 +277,33 @@ function validate(value: SchemaValue, schema: JsfSchema, options: ValidationOpti
   return result
 }
 
-export interface ValidationOptions {
```
lukad, line 275: Moved to `src/validation/schema.ts` to avoid circular dependency between `src/form.ts` and `src/visibility.ts`.

next/src/form.ts
```diff
@@ -26,6 +30,39 @@ export interface ValidationResult {
...
+ * Remove composition keywords and their indices as well as conditional keywords from the path
+ * @param path - The path to clean
+ * @returns The cleaned path
+ * @example
+ * ```ts
+ * cleanErrorPath(['some_object','allOf', 0, 'then', 'field'])
+ * // ['some_object', 'field']
+ * ```
+ */
+function cleanErrorPath(path: ValidationErrorPath): ValidationErrorPath {
```
eng-almeida, line 43: I think we could use a declarative approach for faster readability. Not sure if I missed any corner case here, so feel free to disregard my suggestion if that's the case 👇 

```ts
function cleanErrorPath(path) {
    // Filter out specific keywords and non-string/non-numeric path segments
    return path.filter(segment => 
        typeof segment === 'string' && 
        !['allOf', 'anyOf', 'oneOf', 'then', 'else'].includes(segment)
    );
}
```

Also can we have tests for this function?

next/src/visibility.ts
```diff
@@ -0,0 +1,164 @@
...
+  for (const fieldName in schema.properties) {
+    const fieldSchema = schema.properties[fieldName]
+
+    // Only process object schemas with conditional logic (allOf)
+    if (typeof fieldSchema !== 'object' || fieldSchema === null
+      || Array.isArray(fieldSchema) || !fieldSchema.allOf) {
+      continue
+    }
+
+    const objectField = fields.find(f => f.name === fieldSchema.title || f.name === fieldName)
```
eng-almeida, line 37:
```suggestion
    const objectField = fields.find(field => field.name === fieldSchema.title || field.name === fieldName)
```

next/src/visibility.ts
```diff
@@ -0,0 +1,164 @@
...
+      const ifErrors = validateSchema(values, ruleObj.if!, options)
+      const matches = ifErrors.length === 0
+
+      // Prevent fields from being shown when required fields have type errors
+      let hasTypeErrors = false
+      if (matches
+        && typeof ruleObj.if === 'object'
+        && ruleObj.if !== null
+        && 'required' in ruleObj.if
+        && Array.isArray(ruleObj.if.required)) {
```
eng-almeida, line 97: Do we need both condition? Isn't the `Array.isArray` enough?

```suggestion
        && Array.isArray(ruleObj.if.required)) {
```

next/src/visibility.ts
```diff
@@ -0,0 +1,164 @@
...
+ */
+function isFieldVisible(
+  fieldName: string,
+  branch: JsfSchema,
+  currentVisibility: boolean,
+): boolean {
+  let isVisible = currentVisibility
+
+  if (typeof branch === 'object' && branch !== null) {
+    if ('required' in branch && Array.isArray(branch.required)) {
```
eng-almeida, line 150: Same question as above 🙂 I think both these condition overlap, so I guess we could just have `if (Array.isArray(branch.required))`

eng-almeida (commented):
Hey @lukad, I've juse reviewed the PR and left a few comments/questions. I think the code makes sense overall but I'm not very familiar with the implementation, so take it with a grain of salt 🙂

next/src/visibility.ts
lukad: Great catch, yes the double check here is redundant.
I think even `matches && Array.isArray(ruleObj?.if?.required)` should work? I'll try that.

next/src/field/schema.ts
```diff
@@ -137,6 +137,14 @@ export function buildFieldSchema(
     ...(errorMessage && { errorMessage }),
   }
 
+  if (schema.const) {
+    field.const = schema.const
+
+    if (inputType === 'checkbox') {
```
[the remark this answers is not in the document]
lukad: Good question, I'm not 100% sure about this but I think library should not dictate what inputTypes can be used with it.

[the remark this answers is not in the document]
eng-almeida: > <input> type attribute allows + some custom stuff like 'fieldset'

Yeah there are some exceptions. For example `textArea` is not an `input`or `money` is not a type 😅 Anyway this is not a blocker for this PR. We can discuss this is a follow up PR 🙂

next/src/visibility.ts
lukad: Update:
> I think even `matches && Array.isArray(ruleObj?.if?.required)` should work?

Does not work because we must still verify that `ruleObj.if` is an actual object because it could be a boolean. So I've just removed the redundant `'required' in ruleObj.if` check.

next/src/visibility.ts
lukad: Removed the `'required' in branch`, same as above.

lukad:
Thanks for the comments & questions @eng-almeida, I've just pushed some commits with a bunch of tests. Could you have another look? 🙏

eng-almeida (approved):

next/src/visibility.ts
```diff
@@ -0,0 +1,163 @@
...
+
+      const ifErrors = validateSchema(values, ruleObj.if!, options)
+      const matches = ifErrors.length === 0
+
+      // Prevent fields from being shown when required fields have type errors
+      let hasTypeErrors = false
+      if (matches
+        && typeof ruleObj.if === 'object'
+        && ruleObj.if !== null
+        && Array.isArray(ruleObj.if.required)) {
```
lukad, line 96: This can be added to the filter above in line 84 and we can eliminate the checks above.

next/src/visibility.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @param branch - The schema clause (either 'then' or 'else')
+ * @param currentVisibility - The current visibility state
+ * @returns The updated visibility state
+ *
+ * Visibility logic:
+ * - If the field is in the required array → make visible
+ * - If the field is explicitly false in properties → make hidden
+ * - Otherwise, preserve current visibility
+ */
+function isFieldVisible(
```
lukad, line 141: The way it should work:
* a field is always visible unless the schema is `false`, or in this case the taken if branch.
* when both `const` and `default` are set on a schema it sets forcedValue and hides it but does not set isVisible to false

PR #158, opened 2025-04-14
feat(next): DEVXP-2846: don't output null & undefined radio options

lukad:
* Skips radio options that don't have a `const` value

### DIFF
--- a/next/src/field/type.ts
+++ b/next/src/field/type.ts
@@ -28,3 +28,15 @@ export interface Field {
   // Allow additional properties from x-jsf-presentation (e.g. meta from oneOf/anyOf)
   [key: string]: unknown
 }
+
+/**
+ * Field option
+ * @description
+ * Represents a key/value pair that is used to populate the options for a field.
+ * Will be created from the oneOf/anyOf elements in a schema.
+ */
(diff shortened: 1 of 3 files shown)


lukad:
I noticed that v0 actually only skips radio options with an explicit `"const": null` but generates options where the `const` is absent i.e. `undefined`. However that results in a broken form so I'm also skipping undefined options.

Check out this example in the playground:

```
{
  "properties": {
    "foo": {
      "oneOf": [
        {
          "const": "hello",
          "title": "Hello"
        },
        {
          "const": null,
          "title": "Null"
        },
        {
          "title": "Undefined"
        }
      ],
      "x-jsf-presentation": {
        "inputType": "radio"
      },
      "type": "string"
    }
  }
}
```

The option "Undefined" can not be selected.

CleanShot 2025-04-14 at 14 53 36@2x

antoniocapelo (dismissed):

dragidavid (dismissed):

next/src/field/type.ts
dragidavid, line 1: Good stuff creating this type separately!

dragidavid (approved):

PR #159, opened 2025-04-14
chore: change deprecated `presentation` to `x-jsf-presentation` in test helpers

lukad:
This changes all test helpers to use `x-jsf-presentation` instead of the deprecated `presentation` so that they can be used with the next version.


dragidavid (approved):
TIL 💡

PR #160, opened 2025-04-14
feat(next): Add date validation

antoniocapelo:
# Date Validation Support

## Changes
- Added date validation support with `minDate` and `maxDate` constraints on the `x-jsf-presentation` property
- Added custom date format validation messages
- Added new error types for date validation
- Added test coverage for date validation scenarios

## Example Usage
```typescript
const schema = {
  type: 'string',
  format: 'date',
  'x-jsf-presentation': {
    minDate: '2022-01-01',
    maxDate: '2022-12-31'
  }
}
```

## Error Messages
- Invalid format: "Must be a valid date in yyyy-mm-dd format. e.g. 2024-03-21"
- Min date violation: "The date must be 2022-01-01 or after."
- Max date violation: "The date must be 2022-12-31 or before."

### DIFF
--- a/next/src/errors/messages.ts
+++ b/next/src/errors/messages.ts
@@ -1,6 +1,7 @@
 import type { SchemaValidationErrorType } from '.'
 import type { JsfSchemaType, NonBooleanJsfSchema, SchemaValue } from '../types'
 import { randexp } from 'randexp'
+import { DATE_FORMAT } from '../validation/custom/date'
 
 export function getErrorMessage(
   schema: NonBooleanJsfSchema,
@@ -40,6 +41,12 @@ export function getErrorMessage(
       if (schema.format === 'email') {
         return 'Please enter a valid email address'
       }
+
+      if (schema.format === 'date') {
+        const currentDate = new Date().toISOString().split('T')[0]
+        return `Must be a valid date in ${DATE_FORMAT.toLowerCase()} format. e.g. ${currentDate}`
+      }
+
       return `Must be a valid ${schema.format} format`
     // Number validation
     case 'multipleOf':
--- a/next/src/errors/messages.ts
+++ b/next/src/errors/messages.ts
@@ -52,6 +52,11 @@ export function getErrorMessage(
       return `Must be greater or equal to ${schema.minimum}`
     case 'exclusiveMinimum':
       return `Must be greater than ${schema.exclusiveMinimum}`
+      // Date validation
+    case 'minDate':
+      return `The date must be ${schema['x-jsf-presentation']?.minDate} or after.`
+    case 'maxDate':
+      return `The date must be ${schema['x-jsf-presentation']?.maxDate} or before.`
   }
 }
 
--- a/next/src/validation/custom/date.ts
+++ b/next/src/validation/custom/date.ts
@@ -69,24 +69,21 @@ export function validateDate(
   const isEmptyString = value === ''
   const isUndefined = value === undefined || (value === null && options.treatNullAsUndefined)
   const isEmpty = isEmptyString || isUndefined
+  const errors: ValidationError[] = []
 
   if (!isString || isEmpty || schema['x-jsf-presentation'] === undefined) {
-    return []
+    return errors
   }
 
   const { minDate, maxDate } = schema['x-jsf-presentation']
 
   if (minDate && !validateMinDate(value, minDate)) {
-    return [
-      { path, validation: 'minDate' },
-    ]
+    errors.push({ path, validation: 'minDate' })
   }
 
   if (maxDate && !validateMaxDate(value, maxDate)) {
-    return [
-      { path, validation: 'maxDate' },
-    ]
+    errors.push({ path, validation: 'maxDate' })
   }
 
-  return []
+  return errors
 }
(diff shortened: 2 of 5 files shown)


next/src/validation/custom/date.ts
```diff
@@ -0,0 +1,92 @@
...
+type DateComparisonResult = 'LESSER' | 'GREATER' | 'EQUAL'
+
+/**
+ * Compare two dates and returns the ComparisonResult (LESSER, GREATER, EQUAL)
+ * @param d1 - The first date
+ * @param d2 - The second date
+ * @returns 'LESSER' if the first date is less than the second date, 'GREATER' if the first date is greater than the second date, 'EQUAL' if the two dates are equal
+ */
+function compareDates(d1: string, d2: string): DateComparisonResult {
+  const date1 = new Date(d1).getTime()
```
eng-almeida, line 15: comment: maybe I'm being extra cautios but there's no input validation for d1 or d2. Should we check if d1 and d2 are valid dates?

next/src/validation/custom/date.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/**
+ * Validate that a date value is valid
+ * @param value - The value to validate
+ * @param schema - The schema to validate against
+ * @param options - The validation options
+ * @param path - The path to the value
+ * @returns An array of validation errors
+ * @description This function validates that a dat string value matches the maximum
```
eng-almeida, line 59:
```suggestion
 * @description This function validates that a date string value matches the maximum
```

next/src/validation/custom/date.ts
```diff
@@ -0,0 +1,92 @@
...
+  const isString = typeof value === 'string'
+  const isEmptyString = value === ''
+  const isUndefined = value === undefined || (value === null && options.treatNullAsUndefined)
+  const isEmpty = isEmptyString || isUndefined
+
+  if (!isString || isEmpty || schema['x-jsf-presentation'] === undefined) {
+    return []
+  }
+
+  const { minDate, maxDate } = schema['x-jsf-presentation']
```
eng-almeida, line 77: Not sure if this an a actual use case, but if minDate an maxDate are both defined, we're checking only for minDate first. Should we handle these cases first and return an error message indicating that a date must be between X and Y?

antoniocapelo: Makes sense 👍

next/src/validation/custom/date.ts
lukad: Good point, do you think throwing an exception would be appropriate here? I think returning a `ValidationError` would not be right.

dragidavid: We could technically return a validation error since it is not a valid date after all. 🤔 What does v0 do in this case? Do we handle it?

next/src/validation/custom/date.ts
dragidavid: We can probably start adding TODOs or something similar for things we want to change once we adopt this next version more and more and v0 gets deprecated.

next/src/validation/custom/date.ts
eng-almeida: That's true @antoniocapelo it would mean that users made some mistakes along the way. However I think it wouldn't cause any harm to gracefully handle these cases. Not a blocker though, in case you disagree.

next/src/errors/messages.ts
```diff
@@ -52,6 +59,11 @@ export function getErrorMessage(
       return `Must be greater or equal to ${schema.minimum}`
     case 'exclusiveMinimum':
       return `Must be greater than ${schema.exclusiveMinimum}`
+      // Date validation
+    case 'minDate':
+      return `The date must be ${schema['x-jsf-presentation']?.minDate} or after.`
+    case 'maxDate':
+      return `The date must be ${schema['x-jsf-presentation']?.maxDate} or before.`
```
dragidavid, line 66: Question: can we have a case where `minDate` or `maxDate` does not exist? Should we have a fallback error message for those cases or is that not possible?

next/src/validation/custom/date.ts
lukad: > AFAIK we're only passing down the error message for the first validation error

Correct.

Maybe I misunderstood the suggestion but I understand it like this:
```
if (minDate && maxDate && (!validateMinDate(value, minDate) || !validateMaxDate(value, maxDate))) {
  errors.push({path, validation: 'dateRange'})
} else {
  if (minDate && !validateMinDate(value, minDate)) {
    errors.push({ path, validation: 'minDate' })
  }

  if (maxDate && !validateMaxDate(value, maxDate)) {
    errors.push({ path, validation: 'maxDate' })
  }
}
```

next/src/errors/messages.ts
antoniocapelo: Humm… that's a good question. Currently, the only way a `maxDate` error can exist is **if** a `x-jsf-presentation['maxDate']` exists in the schema, so we're safe. In the future this can change, but I believe we should move fast and don't overthink things

next/src/validation/custom/date.ts
antoniocapelo: I'll keep it as is for now 🙁

next/src/validation/custom/date.ts
antoniocapelo: What I think we could do is solve this with types. We can refine our types so that **if** a `x-jsf-presentation[max/minDate]` is provided, the format should be either `date` or `datetime`. But at this point, I believe we should move faster and leave those improvements for later. Sounds good?

next/src/errors/messages.ts
dragidavid: Agreed 👌

PR #164, opened 2025-04-16
feat(v1): add modify utility

antoniocapelo:
# Add `modify` function for v1

This PR replicates the `modify` utility (that already exists in `v0`) which allows developers to customize JSON schemas programmatically. 

## Dependencies

- Added `lodash` and `@types/lodash` for robust object manipulation

## Implementation Details

Most of the logic was copied from `v0` and ported to TypeScript, including the test suite. However, there were some minor changes:

- added some type guards so that `tsc` considers the code flow valid
- removed the `presentation`/`errorMessage` shorthand — since these keywords are not supported on `v1`, it doesn't make sense to have them at the `modify` level

I decided to do this because (and not have a single file for `modify` across the repo) because v1 will remove support for some keywords, as mentioned above. Instead of having different behaviors based on the version being used, duplicating the code it's acceptable to me, since this code is mostly stable. Having it in TS makes it easier to spot errors BUT for the sake of speed I won't go into setting up TS build in `v0`. With that being said, I believe porting v0 code to TS for the v1 project is the best way to go, for speed and quality purposes.


next/src/modify-schema.ts
```diff
@@ -0,0 +1,429 @@
+import type { JsfSchema } from './types'
```
sandrina-p, line 1: I trusted your copy-pasting skills + tests, so instant approval. LGTM 👍

sandrina-p (approved):

next/src/index.ts
```diff
@@ -1 +1,2 @@
 export { createHeadlessForm, CreateHeadlessFormOptions, ValidationOptions } from './form'
+export { modifySchema as modify } from './modify-schema'
```
sandrina-p, line 2: q: Why calling it `modify-schema.ts`? Why not just modify? Or a folder `modify/index`?

lukad (approved):

next/src/index.ts
antoniocapelo: because we already have a file called `mutations`, for mutating fields. Having a ´modify` and a `mutations` field can be confusing IMO, so I decided to be more specific on what we're modifying. (I kept the export the same though)

PR #165, opened 2025-04-17
feat(next): DEVXP-2540: validate arrays

lukad:
This adds support for validating arrays using all array validation keywords except for `unevaluatedItems` which seems to be a bit more complex is not supported by v0 anyway.

Support for generating group array fields will be added in a separate PR.


next/src/validation/util.ts
dragidavid, line 1: Super nit: Why don't we have these utils in the main [utils.ts](https://github.com/remoteoss/json-schema-form/blob/main/next/src/utils.ts) file?

next/src/validation/array.ts
```diff
@@ -0,0 +1,184 @@
...
+ * - When the `contains` keyword is defined without `minContains` and `maxContains`, the array must contain at least one item that is valid against the `contains` schema.
+ * - When the `contains` keyword is defined with `minContains` and `maxContains`, the array must contain a number of items that is between `minContains` and `maxContains` that are valid against the `contains` schema.
+ */
+function validateArrayContains(
+  value: SchemaValue[],
+  schema: NonBooleanJsfSchema,
+  options: ValidationOptions,
+  path: ValidationErrorPath,
+): ValidationError[] {
+  if (typeof schema.contains !== 'object' || schema.contains === null) {
```
dragidavid, line 133: Q: Should we check if `contains` exists before?

next/src/validation/array.ts
```diff
@@ -0,0 +1,184 @@
...
+      errors.push({ path, validation: 'minContains' })
+    }
+
+    if (schema.maxContains !== undefined && contains > schema.maxContains) {
+      errors.push({ path, validation: 'maxContains' })
+    }
+  }
+
+  return errors
+}
```
dragidavid, line 160: Can there be a case when we have both `minContains` and `maxContains`? I wonder if we should check for that and ensure that `minContains <= maxContains` 🤔

next/src/validation/array.ts
```diff
@@ -0,0 +1,184 @@
...
+    return []
+  }
+
+  for (let i = 0; i < values.length; i++) {
+    for (let j = i + 1; j < values.length; j++) {
+      if (deepEqual(values[i], values[j])) {
+        return [{ path, validation: 'uniqueItems' }]
+      }
+    }
+  }
```
dragidavid, line 182: 💡/Q: Purely as a question.. would it be safe to stringify the value here? 🤔 If we can do that, then we can implement this in a slightly optimized way:
```
  const seen = new Set<string>();

  for (const value of values) {
    const stringValue = JSON.stringify(value);
    if (seen.has(stringValue)) {
      return [{ path, validation: 'uniqueItems' }];
    }
    seen.add(stringValue);
  }
  ```

**Review by dragidavid — COMMENTED** [review]
Left a couple comments but overall this is really good stuff! ⭐

File: next/src/types.ts [hunk 4]
```diff
@@ -8,7 +8,7 @@ export type JsfSchemaType = Exclude<JSONSchema, boolean>['type']
 /**
  * Defines the type of a value in the form that will be validated against the schema.
  */
-export type SchemaValue = string | number | ObjectValue | null | undefined | Array<SchemaValue>
+export type SchemaValue = string | number | ObjectValue | null | undefined | Array<SchemaValue> | boolean
```
lukad, line 11: `boolean` was missing here 🤷 noticed it while writing a unit test.

next/src/validation/array.ts
lukad: I don't think we should handle that case. If you define a schema that has `minContains > maxContains` then you'll get both a `minContains` and a `maxContains` error. If we really wanted to handle this, I guess we would have to either throw an exception or add a special Error type because I don't see this as a `ValidationError`. We don't validate the schema, we validate the value (instance as the spec calls it) against the schema.

next/src/validation/array.ts
lukad: I just tested it and there are two tests from the json schema test suite failing:

```
  ● JSON Schema Test Suite › uniqueItems validation › property order of array of objects is ignored

    expected [ { foo: 'bar', bar: 'foo' }, { bar: 'foo', foo: 'bar' } ] not to be valid for {
      '$schema': 'https://json-schema.org/draft/2020-12/schema',
      uniqueItems: true
    }
```

and

```
  ● JSON Schema Test Suite › uniqueItems validation › objects are non-unique despite key order

    expected [ { a: 1, b: 2 }, { b: 2, a: 1 } ] not to be valid for {
      '$schema': 'https://json-schema.org/draft/2020-12/schema',
      uniqueItems: true
    }
```

next/src/validation/array.ts
lukad: Good point, I had another look at this and realized that only continuing when we have a non-null object here is not even correct. I've changed it to `if (!('contains' in schema)) { return [] }`. Now it works for schemas such as `{"contains": false}` and there were tests in the json schema test suite that I missed for this exact case.

next/src/validation/array.ts
eng-almeida: Ah O(n^2) time complexity 😅 Would something like this work:

```
  const seen = new Map();
  
  for (let i = 0; i < values.length; i++) {
    for (const [j, prevItem] of seen.entries()) {
      if (deepEqual(values[i], prevItem)) {
        return [{ path, validation: 'uniqueItems' }];
      }
    }
    seen.set(i, values[i]);
  }
  ```
  
While this isn't O(n) in the best case (it's still O(n^2) in the worst case), it's probably more efficient 🤔 
- We're only comparing each new item with previously seen items rather than all future items
- We don't need to re-compare items we've already checked against each other
- We bail out as soon as we find a duplicate.

Let me know, if that makes sense

Comment (lukad) [reply next/src/validation/array.ts]: That's a neat solution! Tbh I don't think it will save us a lot of comparisons but it's good have this optimization. Tests are passing 👍

Comment (eng-almeida) [reply next/src/validation/array.ts]: Yeah, this is probably micro performance improvements in our case but still with O(n^2) and deepEqual in the middle, is better to be safe on that front.

File: next/src/validation/array.ts (hunk 1 above)
Comment (dragidavid) [reply next/src/validation/array.ts]: Makes sense! Thanks! 😄

File: next/src/validation/array.ts (hunk 3 above)
Comment (dragidavid) [reply next/src/validation/array.ts]: Yup, I didn't test my suggestion so wasn't sure if the tests would pass. Happy with @eng-almeida's suggestion here 😄

**Review by dragidavid — APPROVED** [review]

File: next/src/validation/array.ts [hunk 5]
```diff
@@ -0,0 +1,188 @@
...
+ * Validates the prefixItems constraint of an array.
+ * If the `prefixItems` keyword is defined, each item in the array must match the schema of the corresponding prefix item.
+ */
+function validatePrefixItems(schema: NonBooleanJsfSchema, values: SchemaValue[], options: ValidationOptions, path: ValidationErrorPath): ValidationError[] {
+  if (!Array.isArray(schema.prefixItems)) {
+    return []
+  }
+
+  const errors: ValidationError[] = []
+  for (const [i, item] of values.entries()) {
```
antoniocapelo, line 107: **Nitpick**

Instead of iterating through all items in the array and checking if the index is less than the prefix items length, we could use the same approach as in `validateItems`:

```ts
  const endIndex =schema.prefixItems.length;

  for (const [i, item] of values.slice(0, endIndex).entries()) {
      errors.push(...validateSchema(item, schema.prefixItems[i] as JsfSchema, options, [...path, 'prefixItems', i]))
  }
```

antoniocapelo (approved):

next/src/validation/array.ts
sandrina-p: > We don't validate the schema, we validate the value (instance as the spec calls it) against the schema.

Exactly. 💯 Never forget this. JSF is a schema validator, not a schema linter. "Impossible schemas" are everywhere, eg `{ type: ["string"], enum: [1, 2, 3] }` it's contradicting and we do not validate it.

next/test/validation/array.test.ts
```diff
@@ -0,0 +1,264 @@
...
+          validation: 'minContains',
+        },
+      ])
+    })
+  })
+
+  describe('maxContains', () => {
+    const schema = {
+      type: 'array',
+      contains: { type: 'number' },
```
sandrina-p, line 247: @lukad incredible work here! I can tell you we had teams asking for the `minItems` and `contains` in the past, and now seeing it implemented, is so cool! ✨

PR #167, opened 2025-04-17
fix: fallback to schema type when presentation is not present

antoniocapelo:
1. **Input Type Handling**:
- Added strict input type validation with `strictInputType` parameter (parity with `v0`)
- New `getInputType` (and `getInputTypeFromSchema`) function to determine field types
- Input types are now strongly typed via `FieldType` enum instead of string

2. **Field Types**:
- Added explicit `FieldType` with values: 'text', 'number', 'select', 'file', 'radio', 'group-array', 'email', 'date', 'checkbox', 'fieldset', 'money', 'country'
- Added "Deprecated note" for `type` field in favor of `inputType` (same as `v0`)

3. **Schema Building**:
- `buildFieldSchema` now accepts `strictInputType` parameter
- Input type determination is more sophisticated, checking:
  - `x-jsf-presentation.inputType` first
  - Falls back to schema-based type inference if not strict
  - Throws error in strict mode if inputType missing

4. **Form Creation**:
- Added `strictInputType` option to `CreateHeadlessFormOptions`
- This propagates through to field building functions

5. **Type Safety**:
- Improved TypeScript types throughout
- JsfPresentation now uses strongly typed `FieldType`
- Better type handling in schema modification functions

### DIFF
--- a/next/src/field/schema.ts
+++ b/next/src/field/schema.ts
@@ -55,6 +55,8 @@ function getInputTypeFromSchema(type: JsfSchemaType, schema: NonBooleanJsfSchema
       }
       return 'select'
     }
+    case 'boolean':
+      return 'checkbox'
     default:
       return 'text'
   }
--- a/next/test/fields.test.ts
+++ b/next/test/fields.test.ts
@@ -341,6 +341,7 @@ describe('fields', () => {
         .toThrow(/Strict error: Missing inputType to field "Test"/)
     })
 
+    // Skipping this test until we have group-array support
     it.skip('defaults to group-array for schema with no type but items.properties', () => {
       const schema = {
         items: {
@@ -435,6 +436,15 @@ describe('fields', () => {
       expect(field?.inputType).toBe('fieldset')
     })
 
+    it('uses checkbox input for boolean type', () => {
+      const schema = {
+        type: 'boolean',
+      }
+      const field = buildFieldSchema(schema, 'test')
+      expect(field?.inputType).toBe('checkbox')
+    })
+
+    // Skipping these tests until we have group-array support
     describe.skip('array type inputs', () => {
       it('uses group-array when items has properties', () => {
         const schema = {


next/src/field/schema.ts
```diff
@@ -121,7 +121,7 @@ export function buildFieldSchema(
   const errorMessage = schema['x-jsf-errorMessage']
 
   // Get input type from presentation or fallback to schema type
-  const inputType = presentation.inputType || 'text'
+  const inputType = presentation.inputType || getJsonType(schema)
```
eng-almeida, line 124: Is it safe to do this? Can't `schema.type` be different than `inputType`? Eg: schema.type can be `string` which is different from what we used to have here - `text`

antoniocapelo: You're right, I now see it 👍 Please check my most recent commit

lukad: So I checked what v0 does and we only fall back to something other than the x-jsf-presentation inputType when the `strictInputType` option is not active. And there is a little bit more fallback logic.

From `src/internals/field.js`. I guess we need to copy that logic for compatibilty.

```
/**
 * @param {object} fieldProperties - any JSON schema field
 * @param {boolean=} strictInputType - From config.strictInputType
 * @param {name=} name - Field id (unique name)
 * @returns {keyof supportedTypes}
 */
export function getInputType(fieldProperties, strictInputType, name) {
  const presentation = pickXKey(fieldProperties, 'presentation') ?? {};
  const presentationInputType = presentation?.inputType;

  if (presentationInputType) {
    return presentationInputType;
  }

  if (strictInputType) {
    throw Error(`Strict error: Missing inputType to field "${name || fieldProperties.title}".
You can fix the json schema or skip this error by calling createHeadlessForm(schema, { strictInputType: false })`);
  }

  if (!fieldProperties.type) {
    if (fieldProperties.items?.properties) {
      return supportedTypes.GROUP_ARRAY;
    }
    if (fieldProperties.properties) {
      return supportedTypes.SELECT;
    }
    return jsonTypeToInputType[jsonTypes.STRING](fieldProperties);
  }

  return jsonTypeToInputType[fieldProperties.type]?.(fieldProperties);
}
```

next/src/field/schema.ts
```diff
@@ -19,6 +20,30 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
...
+
+  if (schemaType === 'string') {
+    return 'text'
+  }
+
+  if (schemaType === 'number') {
+    return 'number'
+  }
+
+  return 'text'
```
eng-almeida, line 44: I'm not sure if not having `x-jsf-presentation` is a common scenario but I feel this is still not 100% safe. I can think of 2 cases:
- when `schemaType` is `boolean`. In this case it should map to checkbox, no?. As it stands it will default to text.
- there might be cases where type is `[string, null]`. Do you want to cover this case and return `text` when it happens?

next/src/field/schema.ts
antoniocapelo: Thanks @lukad , that makes sense. I'll work on it 👍

next/src/field/schema.ts
antoniocapelo: all good points @joaoalmeida-remote 👍

next/src/field/schema.ts
antoniocapelo: @lukad @joaoalmeida-remote I've made another commit, please check 🙏

next/src/field/schema.ts
antoniocapelo: @joaoalmeida-remote feel free to re-review :)

next/src/field/schema.ts
```diff
@@ -19,6 +19,82 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
   return 'text'
 }
 
+/**
+ * Get the presentation input type for a field from a schema type (ported from v0)
```
antoniocapelo, line 23: **Note**

This logic was ported from `v0` (just moved to a function instead of object)

next/src/field/type.ts
```diff
@@ -7,8 +7,9 @@ export interface Field {
   label?: string
   description?: string
   fields?: Field[]
-  type: string
-  inputType: string
+  // @deprecated in favor of inputType,
+  type: FieldType
```
sandrina-p, line 11: suggestion: Could we not return `type` by default? It's deprecated, we want it out of v1.

As internally we need it ourselves, we can pass a config key like `deprecatedFieldAttrType: true` to output it. WDYT?

antoniocapelo: I'm happy to remove it tbh. If we assume this is a deprecation to be done now, I'd not add that config, I'd remove it completely.

antoniocapelo: Still, I'd prefer to remove this on a future PR. It's safer to keep it, for the time being.

next/src/field/schema.ts
```diff
@@ -19,6 +19,82 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
...
+ * @param type - The schema type
+ * @param schema - The non boolean schema of the field
+ * @returns The input type for the field, based schema type. Default to 'text'
+ */
+function getInputTypeFromSchema(type: JsfSchemaType, schema: NonBooleanJsfSchema): FieldType {
+  if (!type) {
+    return 'text'
+  }
+
+  switch (type) {
```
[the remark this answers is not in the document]
antoniocapelo: totally right @joaoalmeida-remote this was an oversight on my side. Just pushed a fix, can you recheck?

eng-almeida (approved):

PR #168, opened 2025-04-17
feat(next): implement file validation

dragidavid:
This PR adds the missing file validation. This is not really related to json schema in general but more just something that we support with jsf. 

The goal was to mimic what we do in v0 and to make the failing tests pass.

### DIFF
--- a/next/src/errors/index.ts
+++ b/next/src/errors/index.ts
@@ -7,24 +7,44 @@ export type SchemaValidationErrorType =
   /**
    * Core validation keywords
    */
-  | 'type' | 'required' | 'valid' | 'const' | 'enum'
+  | 'type'
+  | 'required'
+  | 'valid'
+  | 'const'
+  | 'enum'
   /**
    * Schema composition keywords (allOf, anyOf, oneOf, not)
    * These keywords apply subschemas in a logical manner according to JSON Schema spec
    */
-  | 'anyOf' | 'oneOf' | 'not'
+  | 'anyOf'
+  | 'oneOf'
+  | 'not'
   /**
    * String validation keywords
    */
-  | 'format' | 'minLength' | 'maxLength' | 'pattern'
+  | 'format'
+  | 'minLength'
+  | 'maxLength'
+  | 'pattern'
   /**
    * Number validation keywords
    */
-  | 'multipleOf' | 'maximum' | 'exclusiveMaximum' | 'minimum' | 'exclusiveMinimum'
+  | 'multipleOf'
+  | 'maximum'
+  | 'exclusiveMaximum'
+  | 'minimum'
+  | 'exclusiveMinimum'
   /**
    * Date validation keywords
    */
-  | 'minDate' | 'maxDate'
+  | 'minDate'
+  | 'maxDate'
+  /**
+   * File validation keywords
+   */
+  | 'fileStructure'
+  | 'maxFileSize'
+  | 'accept'
 
 export type ValidationErrorPath = Array<string | number>
 
--- a/next/src/errors/messages.ts
+++ b/next/src/errors/messages.ts
@@ -72,12 +72,12 @@ export function getErrorMessage(
       return 'Not a valid file.'
     case 'maxFileSize': {
       const limitKB = presentation?.maxFileSize
-      const limitMB = typeof limitKB === 'number' ? convertKBToMB(limitKB) : 'N/A'
-      return `File size too large. The limit is ${limitMB} MB.`
+      const limitMB = typeof limitKB === 'number' ? convertKBToMB(limitKB) : undefined
+      return `File size too large.${limitMB ? ` The limit is ${limitMB} MB.` : ''}`
     }
     case 'accept': {
-      const formats = presentation?.accept || 'N/A'
-      return `Unsupported file format. The acceptable formats are ${formats}.`
+      const formats = presentation?.accept
+      return `Unsupported file format.${formats ? ` The acceptable formats are ${formats}.` : ''}`
     }
     // Arrays
     case 'minItems':
--- a/next/src/errors/messages.ts
+++ b/next/src/errors/messages.ts
@@ -1,6 +1,7 @@
 import type { SchemaValidationErrorType } from '.'
 import type { JsfSchemaType, NonBooleanJsfSchema, SchemaValue } from '../types'
 import { randexp } from 'randexp'
+import { convertKBToMB } from '../utils'
 import { DATE_FORMAT } from '../validation/custom/date'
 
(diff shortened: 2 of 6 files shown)


dragidavid:
Leaving this as draft for now since even though file validation works across the board, it causes some regression with the test suite we run against.

Digging..

next/src/errors/index.ts
```diff
@@ -7,24 +7,44 @@ export type SchemaValidationErrorType =
   /**
    * Core validation keywords
    */
-  | 'type' | 'required' | 'valid' | 'const' | 'enum'
+  | 'type'
```
antoniocapelo, line 10: **Question**

Do we need this formatting diffs?

next/src/errors/messages.ts
```diff
@@ -59,11 +61,24 @@ export function getErrorMessage(
...
     case 'maxDate':
-      return `The date must be ${schema['x-jsf-presentation']?.maxDate} or before.`
+      return `The date must be ${presentation?.maxDate} or before.`
+    // File validation
+    case 'fileStructure':
+      return 'Not a valid file.'
+    case 'maxFileSize': {
+      const limitKB = presentation?.maxFileSize
+      const limitMB = typeof limitKB === 'number' ? convertKBToMB(limitKB) : 'N/A'
+      return `File size too large. The limit is ${limitMB} MB.`
```
antoniocapelo, line 75: **Suggestion**

```ts
      const limitMB = typeof limitKB === 'number' ? convertKBToMB(limitKB) : undefined;
      
      return `File size too large.${limitMB ? ` The limit is ${limitMB} MB.`: ''}`
```

antoniocapelo: I'm assuming this won't break any tests since on v0 we don't have this fallback, right?)

We could also use the same logic on `accept` on L77

next/src/validation/file.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (acceptedFormats.length > 0) {
+      // Check if *at least one* file has an accepted format.
+      const isAnyFileFormatAccepted = files.some((file) => {
+        const nameLower = file.name.toLowerCase()
+        const extension = nameLower.includes('.') ? `.${nameLower.split('.').pop()}` : ''
+        return extension !== '' && acceptedFormats.includes(extension)
+      })
+
+      // Fail only if *none* of the files have an accepted format.
+      if (!isAnyFileFormatAccepted) {
```
antoniocapelo, line 80: **Question**

Shouldn't we fail if we have (at least) one that does not match the accepted formats, instead of doing it when none matches the formats?

next/src/validation/schema.ts
```diff
@@ -142,32 +147,38 @@ export function validateSchema(
...
-      return []
-    }
+    return schema ? [] : [{ path, validation: 'valid' }]
   }
 
-  const typeValidationErrors = validateType(value, schema, path)
-  if (typeValidationErrors.length > 0) {
-    return typeValidationErrors
+  let typeValidationErrors: ValidationError[] = []
+  // Skip standard type validation ONLY if inputType is explicitly 'file'
```
antoniocapelo, line 168: **Note**

This logic feels weird… can we maybe refactor it (we can do on a sep PR). I guess we could handle the `null` case inside the `validateType` fn, for example?

next/src/validation/schema.ts
```diff
@@ -142,32 +147,38 @@ export function validateSchema(
...
   }
 
-  // If the schema defines "required", run required checks even when type is undefined.
-  if (
-    schema.required && isObjectValue(value)
-  ) {
+  if (schema.required && isObjectValue(value)) {
     const missingKeys = schema.required.filter((key: string) => {
       const fieldValue = value[key]
+      // Field is considered missing if it's undefined OR
```
antoniocapelo, line 180: thank u for these comments (:

next/src/validation/schema.ts
```diff
@@ -188,6 +204,9 @@ export function validateSchema(
     ...validateArray(value, schema, options, path),
     ...validateString(value, schema, path),
     ...validateNumber(value, schema, path),
+    // File validation (conditionally run)
+    ...(shouldCallValidateFile ? validateFile(value, schema, path) : []),
```
antoniocapelo, line 208: **Comment**

I think we should keep being consistent here. Our approach so far, in the `validate*` functions is to always call them, and make an early return whenever needed. We should do the same here, i.e., call `validateFile` and inside it check for the conditions that would make it return early.

antoniocapelo (changes requested):
Nice job here!
Most of the code looks good and from the testing it's also working nicely. I've left:
- some minor comments, not blockers
- a suggestion for refactoring and simplifying the type validation logic (can be done separately)
- a blocker comment about keeping the consistency on how we call the individual validator functions

next/src/errors/index.ts
dragidavid: Well, we do not **need** them but they kind of happen every time I save that file so I'm assuming it would eventually be checked in. 🤔

next/src/errors/messages.ts
dragidavid: Good point! Changed it 👌

next/src/validation/file.ts
dragidavid: Good point! I made it this way because its what we do in `yupSchema` as well when we check the file formats. We use `.some` on the array of files so if any of them is OK, it is all good. Only fails if none of the files are the correct format.

I'm open to change this but I think invalid files are not included at the time of submission anyways.

next/src/validation/schema.ts
dragidavid: Hmm, I think that makes sense. I can look into this separately! 👍

next/src/validation/schema.ts
dragidavid: Good point 💡 Changed! Makes the `schema` a lot easier to understand to be fair. Let the specific validation files have all the logic. 👌

next/src/errors/index.ts
antoniocapelo: Ah, really? ok then 👍

next/src/validation/schema.ts
antoniocapelo: Yes! 💯

next/src/validation/file.ts
antoniocapelo: Ah, ok, makes sense. So if a input accepts .doc and .pdf, we check if our file extensions matches any of the accepted. It's totally fine, I guess yesterday I was tired at this point :-) good job

antoniocapelo (approved):
Looks good!

PR #169, opened 2025-04-22
feat(next): custom validations with json logic

antoniocapelo:
This PR aims to add [`json-logic` validations](https://json-schema-form.vercel.app/?path=/docs/api-reference-json-logic--docs#x-jsf-logicvalidations) to `v1`. 

**2 things this PR does not include:**
- conditionals validation _inside_ `x-jsf-logic` property (that will come in a future PR)
- `computedValues` ([link](https://json-schema-form.vercel.app/?path=/docs/api-reference-json-logic--docs#x-jsf-logiccomputedvalues)) for `x-jsf-logic` (future PR)

1. Added `json-logic-js` dependency and its types
2. Added JSON Logic validation support with new types and validation logic:

Key Changes:
1. New error type `json-logic` added to validation errors
2. New schema properties:
```typescript
'x-jsf-logic'?: JsonLogicSchema
'x-jsf-logic-validations'?: string[]
'x-jsf-logic-computedAttrs'?: Record<keyof JsfSchema, string>
```

3. New validation system using JSON Logic rules:
- Rules are defined in schema under `x-jsf-logic.validations`
- Each validation can have a custom error message
- Rules are referenced by name in `x-jsf-logic-validations` array

Example Usage (based on types):
```typescript
const schema = {
  "x-jsf-logic": {
    validations: {
      "checkAge": {
        rule: { ">=": [{ var: "age" }, 18] },
        errorMessage: "Must be 18 or older"
      }
    },
    computedValues: {
      "fullName": {
        rule: { "+": [{ var: "firstName" }, " ", { var: "lastName" }] }
      }
    }
  },
  "x-jsf-logic-validations": ["checkAge"],
  "x-jsf-logic-computedAttrs": {
    title: "fullName"
  }
}
```

## Implementation approach

For this feature, I decided that there are 2 things that are needed to run `json-logic` validations that are referenced on a schema's `x-jsf-validations` array:
1. the validations' **definitions** (done at the root schema)
2. the current form values

Based on this, I created the type `JsonLogicBag` that holds this 2 pieces of data (schema and values). When this _bag_ is absent at a `validateSchema` call, it means that it should be populated by the current schema (the root one). This _bag_ is then propagated "down" to any recursive `validateX` calls — hence the number of files changed. This way, we can run a `json-logic` evaluation at a "node level" while referring to the current context (validation definition and form value).

Finally, you'll see some lines of code added because I added types for the json-logic related properties as well as line comments explaining what each property means.

## Questions for v0 devs:
- I noticed that on `v0` implementation we had a `replaceUndefinedValuesWithNulls`. Why was this needed? The function seems to replace `undefined` and `null` values with `NaN`, is this what's intended?


next/src/validation/json-logic.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * (Ported from v0. TODO: check why we need it and if the name is correct)
+ * We removed undefined values in this function as `json-logic` ignores them.
+ * Means we will always check against a value for validations.
+ *
+ * @param {object} values - a set of values from a form
+ * @returns {object} values object without any undefined
+ */
+function replaceUndefinedValuesWithNulls(values: any = {}) {
```
eng-almeida, line 13: I read your TODO comment above but still, do you want to replace `values` type with something more specific?

eng-almeida: Also the function name doesn't seem to match what it's actually doing

antoniocapelo: > I read your TODO comment above but still, do you want to replace values type with something more specific?

TBH I don't know — that's why I'm asking this question (see PR description), I also mention the wrong naming of the fn:

> I noticed that on v0 implementation we had a replaceUndefinedValuesWithNulls. Why was this needed? The function seems to replace undefined and null values with NaN, is this what's intended?

PR #172, opened 2025-04-29
chore: refactor error message interpolation approach

antoniocapelo:
Adds `schema` and `value` to `ValidationError`, allowing for the validators to pass extra context to the next step (message interpolation). It also removes the need for traversing the schema an extra time to calculate the correct context for each error message. 

This is why there are a lot of changes: I had to go through each validator and pass these 2 extra properties and update every test that was asserting on the errors returned by the `validateSchema` call.

## Changes
- Added `schema` and `value` fields to `ValidationError` interface to track the schema and value that caused the validation error
- Removed `getSchemaAndValueAtPath` function since schema and value are now directly available in the error object
- Simplified `addErrorMessages` function to use the schema and value directly from the error object
- Updated all validation functions to include schema and value in their error objects
- Added test utility `errorLike` to correct the validation error assertions and avoid big changes on tets.

### DIFF
--- a/next/src/validation/schema.ts
+++ b/next/src/validation/schema.ts
@@ -107,7 +107,7 @@ function validateType(
   }
 
   // If none of the conditions matched, it's a type error
-  return [{ path, validation: 'type' }]
+  return [{ path, validation: 'type', schema, value }]
 }
 
 /**
@@ -194,7 +194,7 @@ export function validateSchema(
 
   // Handle boolean schemas
   if (typeof schema === 'boolean') {
-    return schema ? [] : [{ path, validation: 'valid' }]
+    return schema ? [] : [{ path, validation: 'valid', schema, value }]
   }
 
   // Check if it is a file input (needed early for null check)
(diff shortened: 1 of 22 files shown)


next/src/validation/schema.ts
```diff
@@ -224,6 +224,8 @@ export function validateSchema(
       errors.push({
         path: [...path, key],
         validation: 'required',
+        schema: schema?.properties?.[key] || schema,
```
lukad, line 227: I don't think `schema?.properties?.[key]` is correct. It should always be `schema` because the `required` is a constraint on the current `schema` object schema and not it's properties.

lukad (changes requested):
Awesome! This will make some things a lot easier 🚀

I left one comment regarding `required` errors.

next/src/validation/schema.ts
antoniocapelo: that is, the logic for the checkbox error message is expecting the schema to match the actual checkbox schema…

lukad: Ah, I see. `getErrorMessage` needs the schema of the property that's required to see if `schema['x-jsf-presentation']?.inputType === 'checkbox'`. Feels a little bit wrong looking at just the validation error itself...
But let's keep it as it is for now. I can't think of a better way to do it right now.

PR #173, opened 2025-04-29
feat(next): support for json-logic (part 2)

antoniocapelo:
## Changes
1. Renamed and improved core functions:
   - `validateJsonLogic` → `validateJsonLogicRules` for better clarity
   - `replaceUndefinedValuesWithNulls` → `replaceUndefinedAndNullValuesWithNaN` which reflects better what the function does

2. Added support for computed attributes:
   - New function `validateJsonLogicComputedAttributes` to handle dynamic schema attributes
   - added dedicated test suite under `json-logic.test.ts`

3. Better type definitions with `SchemaValue` and `ValidationOptions`

### DIFF
--- a/next/src/validation/json-logic.ts
+++ b/next/src/validation/json-logic.ts
@@ -3,6 +3,7 @@ import type { JsonLogicContext, NonBooleanJsfSchema, ObjectValue, SchemaValue } 
 import type { ValidationOptions } from './schema'
 import jsonLogic from 'json-logic-js'
 import { validateSchema } from './schema'
+import { safeDeepClone } from './util'
 
 /**
  * jsonLogic interprets  undefined and null values differently when running comparisons and that creates inconsistent results.
@@ -79,7 +80,7 @@ export function validateJsonLogicComputedAttributes(
   }
 
   // Create a copy of the schema
-  const schemaCopy: NonBooleanJsfSchema = structuredClone(schema)
+  const schemaCopy: NonBooleanJsfSchema = safeDeepClone(schema)
 
   // Remove the computed attributes from the schema
   delete schemaCopy['x-jsf-logic-computedAttrs']
--- a/next/src/validation/json-logic.ts
+++ b/next/src/validation/json-logic.ts
@@ -79,7 +79,7 @@ export function validateJsonLogicComputedAttributes(
   }
 
   // Create a copy of the schema
-  const schemaCopy: NonBooleanJsfSchema = { ...schema }
+  const schemaCopy: NonBooleanJsfSchema = structuredClone(schema)
 
   // Remove the computed attributes from the schema
   delete schemaCopy['x-jsf-logic-computedAttrs']
@@ -94,7 +94,7 @@ export function validateJsonLogicComputedAttributes(
       return
     }
 
-    const result: any = jsonLogic.apply(computedAttributeRule, replaceUndefinedAndNullValuesWithNaN(formValue))
+    const result: any = jsonLogic.apply(computedAttributeRule, replaceUndefinedAndNullValuesWithNaN(formValue as ObjectValue))
 
     if (typeof result === 'undefined') {
       return
@@ -103,5 +103,6 @@ export function validateJsonLogicComputedAttributes(
     schemaCopy[schemaKey as keyof NonBooleanJsfSchema] = result
   })
 
+  // Validate the modified schema
   return validateSchema(values, schemaCopy, options, path, jsonLogicContext)
 }
--- a/next/src/validation/json-logic.ts
+++ b/next/src/validation/json-logic.ts
@@ -5,8 +5,8 @@ import jsonLogic from 'json-logic-js'
 import { validateSchema } from './schema'
 
 /**
- * This function is needed because undefined and null values behave differently on json-logic.apply function.
- * This ensures consistent results.
+ * jsonLogic interprets  undefined and null values differently when running comparisons and that creates inconsistent results.
+ * This function attempts to fix that (ported from v0).
  *
  * @param {object} values - a set of values from a form
  * @returns {object} values object without any undefined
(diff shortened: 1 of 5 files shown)


next/src/validation/json-logic.ts
```diff
@@ -1,28 +1,30 @@
...
- * We removed undefined values in this function as `json-logic` ignores them.
- * Means we will always check against a value for validations.
+ * jsonLogic interprets  undefined and null values differently when running comparisons and that creates inconsistent results.
+ * This function attempts to fix that (ported from v0).
  *
  * @param {object} values - a set of values from a form
  * @returns {object} values object without any undefined
  */
-function replaceUndefinedValuesWithNulls(values: any = {}) {
+function replaceUndefinedAndNullValuesWithNaN(values: any = {}) {
```
dragidavid, line 14: nit: I always spot this `any` here. I know it is kind of hard to type so its all good but it is screaming at me 😅

next/src/validation/json-logic.ts
```diff
@@ -52,3 +54,54 @@ export function validateJsonLogic(
...
+      return
+    }
+
+    const result: any = jsonLogic.apply(computedAttributeRule, replaceUndefinedAndNullValuesWithNaN(formValue))
+
+    if (typeof result === 'undefined') {
+      return
+    }
+
+    schemaCopy[schemaKey as keyof NonBooleanJsfSchema] = result
```
dragidavid, line 103: question: Is it "safe" to just check for result being undefined before assigning?

dragidavid (commented):
Left 2 minor comments. Feel pretty good about this since you have the tests! 💯

next/src/validation/json-logic.ts
eng-almeida: Yeah, maybe we can type as a `Record` at least?

next/src/validation/json-logic.ts
```diff
@@ -52,3 +54,54 @@ export function validateJsonLogic(
...
+): ValidationError[] {
+  const computedAttributes = schema['x-jsf-logic-computedAttrs']
+
+  // if the current schema has no computed attributes, we skip the validation
+  if (!computedAttributes || Object.keys(computedAttributes).length === 0) {
+    return []
+  }
+
+  // Create a copy of the schema
+  const schemaCopy: NonBooleanJsfSchema = { ...schema }
```
eng-almeida, line 82: Probably not an issue in this particular case but just want to highlight that this only creates a shallow copy, so nested objects will still reference the original schema.

lukad (dismissed):
Nothing to add, it's easy to read and it's clear what the code does. 🚀

next/src/validation/json-logic.ts
```diff
@@ -1,28 +1,30 @@
...
+ * jsonLogic interprets  undefined and null values differently when running comparisons and that creates inconsistent results.
+ * This function attempts to fix that (ported from v0).
  *
  * @param {object} values - a set of values from a form
  * @returns {object} values object without any undefined
  */
-function replaceUndefinedValuesWithNulls(values: any = {}) {
+function replaceUndefinedAndNullValuesWithNaN(values: any = {}) {
   return Object.entries(values).reduce((prev, [key, value]) => {
     return { ...prev, [key]: value === undefined || value === null ? Number.NaN : value }
```
eng-almeida, line 16: suggestion: I think this function would be readable without the reducer and instead we use direct object assignment:

```ts
function replaceUndefinedAndNullValuesWithNaN(values: any = {}) {
  const result = {};
  for (const [key, value] of Object.entries(values)) {
    result[key] = value === undefined || value === null ? Number.NaN : value;
  }
  return result;
}
```

next/src/validation/json-logic.ts
antoniocapelo: That's true, I'm aware of that. I did this because at this point I'm hoping we're at the "leaf" schema level, so nested objects would be rare. I can do a deep clone though, it's probably safer.

next/src/validation/json-logic.ts
antoniocapelo: @dragidavid isn't that what I'm doing on L99?

```ts
    const result: any = jsonLogic.apply(computedAttributeRule, replaceUndefinedAndNullValuesWithNaN(formValue))

    if (typeof result === 'undefined') {
      return
    }
```

next/src/validation/json-logic.ts
antoniocapelo: Sounds good to me 👍

antoniocapelo:
@lukad @joaoalmeida-remote @dragidavid I realized that for full v0 compatibility I still need to implement 2 things:
- throw errors in case of invalid schemas on the json-logic "part" (does this make sense? LMK what you think)
- allow for interpolations inside "x-jsf-errorMessage" nodes, like

```"x-jsf-errorMessage": {
    "minimum": "Must be at least {{minimum_pto}} days"
}
```

next/src/validation/json-logic.ts
dragidavid: Yes! I was wondering if there is anything that we should check for here but after going through it one more time, I take that back! Looks good!

next/src/validation/util.ts
```diff
@@ -68,3 +68,31 @@ export function deepEqual(a: SchemaValue, b: SchemaValue): boolean {
...
 }
+
+/**
+ * Deep clones an object using structuredClone if available, otherwise falls back to JSON.parse/stringify approach.
+ *
+ * @param obj - The object to clone
+ * @returns deep clone of the original object
+ * @throws If the object contains circular references and structuredClone is not available
+ */
+export function safeDeepClone<T>(obj: T): T {
```
dragidavid, line 79: Nice! 🙌

dragidavid (approved):
Went through this again + the draft you linked. All good imo, really easy to understand with the comments in both PRs

PR #174, opened 2025-04-30
fix(v1): use correct checkboxValue on checkbox inputs

antoniocapelo:
Our previous implementation had a bug:
- it only populated `checkboxValue` if the schema had a `const` and had `checkbox` as its input type

This is a problem because it's not taking into account the `type` of the schema. In other words, to mimic v0's behavior, one should set the `checkboxValue` to either what's defined in the `const` property OR to boolean, in case the `type` is `boolean`.

Added 2 groups of tests
- one for boolean validation (not boolean schemas, but schemas of type boolean)
- one for the `fields` generation

### DIFF
--- a/next/src/field/schema.ts
+++ b/next/src/field/schema.ts
@@ -20,6 +20,23 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
 }
 
 /**
+ * Add checkbox attributes to a field
+ * @param inputType - The input type of the field
+ * @param field - The field to add the attributes to
+ * @param schema - The schema of the field
+ */
+function addCheckboxAttributes(inputType: string, field: Field, schema: NonBooleanJsfSchema) {
+  // The checkboxValue attribute indicates which is the valid value a checkbox can have (for example "acknowledge", or `true`)
+  // So, we set it to what's specified in the schema (if any)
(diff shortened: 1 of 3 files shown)


lukad (dismissed):

next/src/field/schema.ts
```diff
@@ -19,6 +19,23 @@ function getJsonType(schema: NonBooleanJsfSchema): string {
...
+ * @param inputType - The input type of the field
+ * @param field - The field to add the attributes to
+ * @param schema - The schema of the field
+ */
+function addCheckboxAttributes(inputType: string, field: Field, schema: NonBooleanJsfSchema) {
+  // The checkboxValue attribute indicates which is the valid value a checkbox can have (for example "acknowledge", or `true`)
+  // So, we set it to what's specified in the schema (if any)
+  field.checkboxValue = schema.const
+
+  // However, if the schema type is boolean, we should the valid value as `true`
```
dragidavid, line 33:
```suggestion
  // However, if the schema type is boolean, we should set the valid value as `true`
```

dragidavid (changes requested):
Just 1 typo. Makes sense otherwise!

PR #176, opened 2025-05-02
feat: simplify custom error messages calculation and allow for interp…

antoniocapelo:
## What

This PR adds the ability for computed error messages (`x-jsf-errorMessage` inside `x-jsf-computedAttrs`) to reference  
computed values of `x-jsf-logic`. For example:

```json
{
    "properties": {
      "some_property": {
        ...
        "x-jsf-logic-computedAttrs": {
          "minimum": "some_computation",
          "x-jsf-errorMessage": {
            "minimum": "Must be at least {{some_computation}} units"
          }
        }
      }
    },
    "required": [
      "some_property"
    ],
    "x-jsf-logic": {
      "computedValues": {
        "some_computation": {
          "rule": {
            "*": [
              {
                "var": "some_property"
              },
              0.5
            ]
          }
        }
      }
    }
  }
```

## How

This is mainly done with 2 changes:
- When applying the custom error message, we don't traverse the schema searching for the node that holds the `x-jsf-errorMessage` values. This has 2 problems:
  - 1. it traverses the tree again (this traversal already happened in the validation phase)
  - 2. it lacks the json logic context to apply the interpolated values
  - Instead, we now use the `schema` that comes attached to the `ValidationError`. This schema already brings any interpolations executed and makes the code much simpler
 - When applying the json logic computed attributes, we check if the attribute we're computing is `x-jsf-errorMessage` and, in that case, we run an `interpolate` function that replaced any handlebar symbols with the respective computed value from the json logic.

### DIFF
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -1,6 +1,6 @@
 import type { ValidationError, ValidationErrorPath } from './errors'
 import type { Field } from './field/type'
-import type { JsfObjectSchema, JsfSchema, NonBooleanJsfSchema, SchemaValue } from './types'
+import type { JsfObjectSchema, JsfSchema, SchemaValue } from './types'
 import type { ValidationOptions } from './validation/schema'
 import { getErrorMessage } from './errors/messages'
 import { buildFieldObject } from './field/object'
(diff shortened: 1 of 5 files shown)


lukad (approved):
This is great! The `ValidationError` change pays off again.

dragidavid (approved):
@antoniocapelo just casually simplifying the whole thing over here 💯

PR #177, opened 2025-05-02
feat(next): group-array field support

lukad:
This adds support for generating fields for `{"type":"array"}` schemas which includes:
* group-arrays (an array schema where the `items` subschema is an `object` schema)
* array schemas where the `items` subschema is anything else e.g. `{"type": "array", "items": {"type": "string}}`


next/src/field/array.ts
```diff
@@ -0,0 +1,62 @@
...
+    fields,
+    isVisible: true,
+  }
+
+  if (schema.title !== undefined) {
+    field.label = schema.title
+  }
+
+  if (schema.description !== undefined) {
+    field.description = schema.description
```
sandrina-p, line 58: bug: We should spread all schema keys (except `items`) down to `field`, why did you decide to not do it? 
For example, if I pass `x-foo`, it will be ignored by JSF.

lukad: Good catch, that was not intentional. I've just pushed a change that reuses the same logic for propagating schema props to fields while excluding certain keywords.

sandrina-p (changes requested):

next/src/field/schema.ts
```diff
@@ -169,6 +173,80 @@ function getFieldOptions(schema: NonBooleanJsfSchema) {
...
+
+  for (const key in schema.properties) {
+    const isRequired = schema.required?.includes(key) || false
+    const field = buildFieldSchema(schema.properties[key], key, isRequired, strictInputType)
+    if (field) {
+      fields.push(field)
+    }
+  }
+
+  const orderedFields = setCustomOrder({ fields, schema })
```
eng-almeida, line 193: comment: what's the rational for the function signature? In some places the function accepts multiple parameters and in another places (like this) uses an object as a parameter. 
This is not a big deal not a blocker either but if we can standardize it in the future, it would be good IMO.

next/src/field/schema.ts
```diff
@@ -190,38 +268,43 @@ export function buildFieldSchema(
...
+      jsonType: 'boolean',
+      required,
+      isVisible: false,
+    }
   }
 
-  if (schema.type === 'array') {
-    throw new TypeError('Array type is not yet supported')
+  // If schema is any other boolean (true), just return null
+  if (typeof schema === 'boolean') {
```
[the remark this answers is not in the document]
lukad: `{ "type": "object", "properties": { "alwaysTrue": true } }` is a valid schema and at some point this function would be called with the subschema for the `alwaysTrue` property which is just `true`. Not sure if there is ever a use for `true` schema but is something we need to handle.

next/src/form.ts
```diff
@@ -94,19 +95,37 @@ function validationErrorsToFormErrors(errors: ValidationErrorWithMessage[]): For
...
+        const [arrayName, _, arrayIndex] = cleanedPath
+
+        if (!(arrayName in result) || typeof result[arrayName] === 'string') {
+          result[arrayName] = [] as Array<null | FormErrors>
+        }
+
+        const array = result[arrayName] as Array<null | FormErrors>
+
+        // Create a new object for this index if it doesn't exist
+        if (array[arrayIndex] === null || array[arrayIndex] === undefined) {
```
eng-almeida, line 113: nitpick: if you prefer a squeezed version of this line, you can use `array[arrayIndex] == null`

next/src/form.ts
```diff
@@ -77,13 +78,13 @@ function cleanErrorPath(path: ValidationErrorPath): ValidationErrorPath {
  * Schema-level error
  * { '': 'The value must match at least one schema' }
  */
-function validationErrorsToFormErrors(errors: ValidationErrorWithMessage[]): FormErrors | null {
+function validationErrorsToFormErrors(errors: ValidationErrorWithMessage[], _value: SchemaValue = {}): FormErrors | null {
```
eng-almeida, line 81: q: what's `_value` for?

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,653 @@
...
+        },
+      },
+    }
+
+    const rootField = buildFieldSchema(schema, 'root', true)
+    const arrayField = rootField?.fields?.[0]
+
+    expect(arrayField).toBeDefined()
+    expect(arrayField?.type).toBe('group-array')
+    expect(arrayField?.required).toBe(true)
```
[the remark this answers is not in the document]
lukad: The organization of tests that I have in mind for this is that everything purely json-schema validation related goes into `tests/validation` and everything fields related into `test/fields`. The other tests I've put into this file are test that check if we map validation errors correctly to `formErrors` returned by `handleValidation`. Does that make sense? Maybe I could add a describe block around these tests to make it clear?

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,653 @@
...
+    expect(field?.type).toBe('group-array')
+
+    const fields = field!.fields!
+    expect(fields).toHaveLength(2)
+    expect(fields[0].name).toBe('name')
+    expect(fields[0].type).toBe('text')
+    expect(fields[0].required).toBe(true)
+    expect(fields[1].name).toBe('age')
+    expect(fields[1].type).toBe('number')
+    expect(fields[1].required).toBe(false)
```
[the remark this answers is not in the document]
lukad: Agreed. I'm rewriting all tests to `toMatchObject`/`toEqual` where it makes sense.

next/src/field/schema.ts
lukad: Didn't write that function but I prefer multiple params tbh and we use that everywhere else. I'll change this one to match the rest.

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,653 @@
...
+    expect(field).toBeDefined()
+    expect(field?.type).toBe('group-array')
+  })
+
+  it('should handle required arrays', () => {
+    const schema: JsfSchema = {
+      type: 'object',
+      required: ['items'],
+      properties: {
+        items: {
```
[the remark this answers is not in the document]
lukad: I've renamed the property to `arrayProperty`.

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,653 @@
...
+      },
+    }
+
+    const field = buildFieldSchema(schema, 'root', true)
+
+    expect(field).toBeDefined()
+    expect(field?.type).toBe('group-array')
+  })
+
+  it('should handle required arrays', () => {
```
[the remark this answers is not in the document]
lukad: I would rather keep them separate. `should handle nested group-arrays` is only testing nesting ability and this one is testing that `required` works.

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,720 @@
...
+    }).formErrors).toEqual({
+      people: [
+        { lastName: 'Required field' },
+        { firstName: 'Required field' },
+        undefined,
+      ],
+    })
+  })
+
+  it.skip('handles validation of nested group-arrays', () => {
```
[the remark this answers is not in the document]
lukad: This one is now enabled. It actually revealed a bug with how formErrors looked like for nested group-arrays which I fixed.

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,653 @@
...
+      type: 'array',
+      items: {
+        type: 'object',
+      },
+    }
+
+    const field = buildFieldSchema(schema, 'root', true)
+
+    expect(field).toBeDefined()
+    expect(field?.type).toBe('group-array')
```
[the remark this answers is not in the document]
lukad: This is now covered by matching on the whole field.

next/test/fields/array.test.ts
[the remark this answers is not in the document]
lukad: I've added that describe block.

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,653 @@
...
+    }).formErrors).toEqual({
+      contacts: [
+        undefined,
+        { phone: 'Phone must be in format: 123-456-7890' },
+        { zipCode: 'Invalid zip code format' },
+      ],
+    })
+  })
+
+  it('handles mixed arrays with different types of validation errors', () => {
```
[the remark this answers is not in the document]
lukad: I'm actually considering deleting this test as it's kind of misleading. Mutating group-array fields doesn't really work and will need a different API as we discussed internally.

next/test/fields/array.test.ts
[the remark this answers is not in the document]
sandrina-p: It does make sense! Some parts are redundant, but this is a case to say ~~less is more~~ more is better :)

next/src/field/schema.ts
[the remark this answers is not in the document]
sandrina-p: At Remote no, never. I don't even know how we would translate that into "fields". But all good for now.

sandrina-p (dismissed):
Fantastic work @lukad! Everything seems to work fine, I'm still testing against the internal MR, but so far, so good! Let's release anoter dev to run against E2E and then we can merge this ⚡

next/src/custom/order.ts
```diff
@@ -1,19 +1,14 @@
 import type { Field } from '../field/type'
 import type { JsfSchema } from '../types'
```
sandrina-p, line 3: scout: Can you update `mockGroupArrayInput` to look better with JSF v1 in the Playground? We need to replace `'x-jsf-presentation'.options` with `oneOf`? Otherwise the radio options render without a title/label

cc @dragidavid

next/src/custom/order.ts
```diff
@@ -1,19 +1,14 @@
...
 
-function sort(params: {
-  fields: Field[]
-  order: string[]
-}): Field[] {
-  const { fields: prevFields, order } = params
-
+function sort(fields: Field[], order: string[]): Field[] {
```
sandrina-p, line 4: bug: I noticed now that `x-jsf-order` is not working. In the screenshot below, the "full_name" should come first, but it's not.

<details>
image
</details>

dragidavid: Yup, that was mentioned before on one of the internal Slack threads. 🫠 At the time I fixed that with the function we use for objects so that should work here again. The code has changed since then though so I discarded it.

next/test/fields/array.test.ts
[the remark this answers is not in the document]
sandrina-p: Call me picky, but it could be better 😅 `arrayProperty` is better than `items`, but not as intuitive as the other friendly examples you created, `tasks`, `departments`, etc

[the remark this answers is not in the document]
lukad: Fair point, I tried to make this one more tangible.

next/test/fields/array.test.ts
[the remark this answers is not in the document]
lukad: That's what I've done now. I've replaced this with a `describe.skip` with two tests that explain why mutation in group-array fields does not work.

next/test/fields/array.test.ts
```diff
@@ -46,6 +46,65 @@ describe('buildFieldArray', () => {
...
+          second_item: { type: 'string' },
+          third_item: { type: 'string' },
+        },
+        'x-jsf-order': ['second_item', 'first_item', 'third_item'],
+      },
+    }
+
+    const field = buildFieldSchema(schema, 'root', true)
+
+    expect(field).toEqual({
```
sandrina-p, line 68: No need for all of this, a simple assertion to the order is enough, no?

expect(fields[0].fields.map(f => f.name)).toEqual(['second_item', 'first_item', 'third_item'])

next/src/custom/order.ts
sandrina-p: @lukad I think you forgot this one 😢

next/test/fields/array.test.ts
lukad: Indeed! That makes a lot more sense.

next/src/custom/order.ts
lukad: Ah yes, sorry. Is there something we need to do int jsf for this?

next/test/fields/array.test.ts
```diff
@@ -0,0 +1,888 @@
...
+          { zipCode: 'Invalid zip code format' },
+        ],
+      })
+    })
+  })
+
+  // These do not work with the current group-array API where all groups share the same `fields` property which
+  // makes it impossible to have different fields for each item in the array.
+  // This applies to all kinds of mutations such as conditional rendering, default values, etc. and not just titles.
+  describe.skip('mutation of array items', () => {
```
sandrina-p, line 789:
```suggestion
  // TODO: Check internal ticket: https://linear.app/remote/issue/RMT-1616/grouparray-hide-conditional-fields
  describe.skip('mutation of array items', () => {
```

next/src/errors/messages.ts
```diff
@@ -81,17 +81,17 @@ export function getErrorMessage(
     }
     // Arrays
     case 'minItems':
-      throw new Error('Array support is not implemented yet')
```
sandrina-p, line 84: todo: We forgot something in these error messages, based [on this convo](https://linear.app/remote/issue/RMT-51/jsf-support-group-array-to-have-visibility-for-conditional-properties#comment-686a7c8d). So although we don't support hiding a conditional field, at least we show a better error message. "Not allowed" is more intuitive than "Always fails", would you agree?

```diff
    case 'valid':
-      return 'Always fails'
-      return 'Not allowed'
```

P.S. This case being called `valid` is so weird, A better name could be `forbidden` or something, as this is a direct match of a schema like `{ address: false }`

next/src/custom/order.ts
dragidavid: Ah its the `enum`. I guess that should be `oneOf` here.

```
          "sex": {
            "description": "We know sex is non-binary but for insurance and payroll purposes, we need to collect this information.",
            "enum": [
              "female",
              "male"
            ],
            "x-jsf-presentation": {
              "inputType": "radio",
              "options": [
                {
                  "label": "Male",
                  "value": "male"
                },
                {
                  "label": "Female",
                  "value": "female"
                }
              ]
            },
            "title": "Child Sex",
            "type": "string"
          }
```

next/src/errors/messages.ts
lukad: 💯 I've changed it now.

next/src/custom/order.ts
dragidavid: Pushed the fix for it :)

next/src/custom/order.ts
lukad: @dragidavid just pushed 305d7cf803789d3792783fe1eedf67118d081558 to this branch updating mockGroupArrayInput as mentioned in the other thread. So resolving this one.

next/src/errors/messages.ts
lukad: > Why do you think forbidden makes assumption about the presentation ui?

You know what? I have no good argument now that I think about it 😅 . Let's do `"forbidden"`. It does make sense, when a value is provided for this schema it fails.

dragidavid:
To explain the last 2 commits:
- I released a new `dev` version for `v0` as we needed the schema updates but the version should remain on the `beta` as it was before

dragidavid (approved):
Did a final test with the new `dev` versions and it seems to be all good! Approving ✅

PR #178, opened 2025-05-02
feat: throw on invalid json-logic schemas

antoniocapelo:
Small PR, ensuring our json-logic module throws Errors at the same point the v0 version does, updating tests accordingly.


dragidavid (approved):
Seems straightforward, adding some errors. 👌

PR #180, opened 2025-05-02
fix: Fields accept any key starting with x-*

sandrina-p:
Given a JSON Schema with a property that has a any keys starting with `x-*`, it passes to the output `fields`, example:

```js
{
  properties: {
    country: {
      title: 'Country',
      type: 'string',
      'x-foo': 123,
    }
  }
}
```

Output: `form.fields[0]["'x-foo"]` exists!

— Ref: [internal ticket](https://linear.app/remote/issue/RMT-577/#comment-4c8652ca) | [Internal E2E tests](https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/40751)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.13-dev.20250502172456",
+  "version": "0.11.12-beta.0",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.12-beta.0",
+  "version": "0.11.13-dev.20250502172456",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


src/helpers.js
```diff
@@ -603,6 +603,9 @@ export function extractParametersFromNode(schemaNode) {
       anyOf: node.anyOf,
       allOf: node.allOf,
       errorMessage,
+      // Used for async selects
+      'x-rmt-enum-async': node['x-rmt-enum-async'],
+      // ...xKeys,
```
sandrina-p, line 608: todo: Let me try to spread all the `x-` keys insted, to avoid more workarounds like this in the future

sandrina-p: Done in the last commit. PR description updated ✅

dragidavid:
Just had a look at this together with the internal ticket. Simple fix as it seems!

Would approve but CI fails so I'll look out for the fix 🙂

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.11.12-beta.0",
+  "version": "0.11.13-dev.20250502172456",
```
sandrina-p, line 3: @dragidavid to answer your comment above: [CI failing](https://github.com/remoteoss/json-schema-form/actions/runs/14800379574/job/41557512457?pr=180) as expected because of this change to the package.json. It's a reminder to revert it before merging the PR (and after the internal E2Es passes). Doing that now with the script `version_as_main`

sandrina-p: Done! [6e55485](https://github.com/remoteoss/json-schema-form/pull/180/commits/6e554852d73361ea665ecfddfb8edb9f344e4f67) can you give a final review?

lukad (approved):

PR #185, opened 2025-05-08
chore(next): Field types

eng-almeida:
Adds TS types for individual fields.

### DIFF
--- a/next/src/field/type.ts
+++ b/next/src/field/type.ts
@@ -1,47 +1,123 @@
 import type { JsfSchemaType } from '../types'
 
-/**
- * WIP type for UI field output that allows for all `x-jsf-presentation` properties to be splatted
- * TODO/QUESTION: what are the required fields for a field? what are the things we want to deprecate, if any?
- */
-export interface Field {
-  name: string
-  label?: string
-  description?: string
-  fields?: Field[]
-  // @deprecated in favor of inputType,
+export type FieldType = 'text' | 'number' | 'select' | 'file' | 'radio' | 'group-array' | 'email' | 'date' | 'checkbox' | 'fieldset' | 'money' | 'country' | 'textarea'
+
+interface BaseField {
   type: FieldType
-  inputType: FieldType
+  name: string
+  label: string
   required: boolean
+  inputType: FieldType
   jsonType: JsfSchemaType
-  isVisible: boolean
-  accept?: string
-  errorMessage?: Record<string, string>
   computedAttributes?: Record<string, unknown>
+  errorMessage: Record<string, string>
+  schema: any
+  isVisible: boolean
+  description?: string
+  statement?: {
+    title: string
+    inputType: 'statement'
+    severity: 'warning' | 'error' | 'info'
+  }
+  [key: string]: unknown
+}
+
+export interface FieldOption {
+  label: string
+  value: string
+  description?: string
+}
+
+export interface FieldSelect extends BaseField {
+  type: 'select'
+  options: FieldOption[]
+}
+
+export interface FieldTextarea extends BaseField {
+  type: 'textarea'
+  maxLength: number
+  minLength?: number
+}
+
+export interface FieldDate extends BaseField {
+  type: 'date'
+  format: string
   minDate?: string
   maxDate?: string
   maxLength?: number
-  maxFileSize?: number
-  format?: string
-  anyOf?: unknown[]
-  options?: unknown[]
-  const?: unknown
-  checkboxValue?: unknown
-
-  // Allow additional properties from x-jsf-presentation (e.g. meta from oneOf/anyOf)
-  [key: string]: unknown
 }
 
-/**
- * Field option
- * @description
- * Represents a key/value pair that is used to populate the options for a field.
- * Will be created from the oneOf/anyOf elements in a schema.
- */
-export interface FieldOption {
+export interface FieldText extends BaseField {
+  type: 'text'
+  maxLength: number
+  maskSecret?: number
+}
+
+export interface FieldRadio extends BaseField {
+  type: 'radio'
+  options: FieldOption[]
+  direction?: 'row' | 'column'
+  const?: string
+}
+
+export interface FieldNumber extends BaseField {
+  type: 'number'
+}
+
+export interface FieldMoney extends BaseField {
+  type: 'money'
+  currency: string
+}
+
+export interface FieldCheckbox extends BaseField {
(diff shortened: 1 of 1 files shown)


eng-almeida:
Hi folks! I've started adding TS types to fields, but I'll need your help to confirm what properties I'm missing, identify common properties that I might be missing as well, what properties and optional/mandatory etc. This is why the PR is still in draft 🙂

next/src/field/type.ts
```diff
@@ -1,47 +1,123 @@
...
+}
+
+export interface FieldSelect extends BaseField {
+  type: 'select'
+  options: FieldOption[]
+}
+
+export interface FieldTextarea extends BaseField {
+  type: 'textarea'
+  maxLength: number
```
lukad, line 38: `maxLength` ia not mandatory in a JsonSchema and if not provided it won't be in the field.

```suggestion
  maxLength?: number
```

next/src/field/type.ts
```diff
@@ -1,47 +1,123 @@
...
-/**
- * Field option
- * @description
- * Represents a key/value pair that is used to populate the options for a field.
- * Will be created from the oneOf/anyOf elements in a schema.
- */
-export interface FieldOption {
+export interface FieldText extends BaseField {
+  type: 'text'
+  maxLength: number
```
lukad, line 52: Same as with `FieldTextArea`.

lukad (commented):
Hey @eng-almeida, I added comments in two places regarding non-optional properties but this basically applies to all non-optional properties on those field types.

We don't actually have any sort of validation that checks that when for example `'x-js-presentation': { 'inputType': 'money'}` is given there is also `'x-js-presentation': { 'currency': 'EUR'}`.

@sandrina-p, is that something we do in v0?

next/src/field/type.ts
```diff
@@ -1,47 +1,130 @@
...
+  errorMessage: Record<string, string>
+  schema: any
+  isVisible: boolean
+  description?: string
+  statement?: {
+    title: string
+    inputType: 'statement'
+    severity: 'warning' | 'error' | 'info'
+  }
+  [key: string]: unknown
```
eng-almeida, line 22: This line _hides_ a lot of TS errors that are hard to address without it. I'll keep investigating...

next/src/field/type.ts
```diff
@@ -1,47 +1,164 @@
...
- * WIP type for UI field output that allows for all `x-jsf-presentation` properties to be splatted
- * TODO/QUESTION: what are the required fields for a field? what are the things we want to deprecate, if any?
- */
-export interface Field {
-  name: string
-  label?: string
-  description?: string
-  fields?: Field[]
-  // @deprecated in favor of inputType,
+export type FieldType = 'text' | 'number' | 'select' | 'file' | 'radio' | 'group-array' | 'email' | 'date' | 'checkbox' | 'fieldset' | 'money' | 'country' | 'textarea'
```
sandrina-p, line 3: todo: Here's my controversial opinion: JSF does not care about `FieldType`, it's whatever it's on the JSON schema. We do not provide any extra logic/validation just based on it, right? (right? 👀)

So answering @lukad question:

> We don't actually have any sort of validation that checks that when for example 'x-js-presentation': { 'inputType': 'money'} is given there is also 'x-js-presentation': { 'currency': 'EUR'}. @sandrina-p, is that something we do in v0?

We don't and we shouldn't. That logic is a concern of Remote internals, not JSF as headless generator/validator. If we ever validate that, it's another layer on top of JSON-SCHEMA-FORM

Does it make sense to you both?

next/src/field/type.ts
```diff
@@ -1,47 +1,164 @@
...
   isVisible: boolean
-  accept?: string
-  errorMessage?: Record<string, string>
-  computedAttributes?: Record<string, unknown>
+}
+
+export interface FieldOption {
+  label: string
+  value: string
+  description?: string
```
sandrina-p, line 20: todo:
1. The `FieldOption` is a mirror of `oneOf`, so if the `oneOf` has, `foo` inside, then the Type includes `foo` too.
2. `value` is not necessarily a string, can be anything, number, bool, even object. It's a mirror of `oneOf.const` 🪩
3. `value` can be optional too :p Look [at this example with `pattern`](https://json-schema-form.vercel.app/?path=/story/demos-fields--select-allow-create)

next/src/field/type.ts
sandrina-p: P.S. I know, this kind affects most of your MR, as the field types should be based on the json schema `type`, not the `inputType` 😶

next/src/field/type.ts
```diff
@@ -1,47 +1,164 @@
...
- * Field option
- * @description
- * Represents a key/value pair that is used to populate the options for a field.
- * Will be created from the oneOf/anyOf elements in a schema.
- */
-export interface FieldOption {
+export interface FieldRadio extends BaseField {
+  type: 'radio'
+  options: FieldOption[]
+  direction?: 'row' | 'column'
```
sandrina-p, line 52: todo: Here's another example. This is Remote needs, not a JSF concern. A few other examples down the line (currency, fileDownload, etc... All of those types can exist, but not at JSF.

next/src/field/type.ts
```diff
@@ -1,47 +1,164 @@
...
+}
+
+export interface GroupArrayField extends BaseField {
+  type: 'group-array'
+  name: string
   label: string
-  value: unknown
-  [key: string]: unknown
+  description: string
+  fields: () => Field[]
```
sandrina-p, line 102: bug: This is v0, in v1 is `fields: Field[]` (Once https://github.com/remoteoss/json-schema-form/pull/177 is merged)) :D

next/src/field/type.ts
eng-almeida: You're right 💯 ! Probably the right move is to extend these types on Remote SDK project which uses Remote JSON Schemas 👍

next/src/field/type.ts
```diff
@@ -1,47 +1,123 @@
 import type { JsfSchemaType } from '../types'
```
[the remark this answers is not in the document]
eng-almeida: Hey @sandrina-p, sorry for dropping the ball here. I'll try to wrap this up next week 🤞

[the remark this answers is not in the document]
sandrina-p: Hi @eng-almeida! Could you wrap this up this year? 🎁

PR #186, opened 2025-05-12
fix(next) - fix types property in package.json

gabrielseco:
I started to implement the package in remote-flows but found typescript errors when installing the package.

```
Could not find a declaration file for module '@remoteoss/json-schema-form'. '/Users/gabriel/remote/remote-flows/node_modules/@remoteoss/json-schema-form/dist/index.mjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/remoteoss__json-schema-form` if it exists or add a new declaration (.d.ts) file containing `declare module '@remoteoss/json-schema-form';`
```

My assumption is that my other library doesn't find the types


lukad (dismissed):
Thanks @gabrielseco!

next/package-lock.json
gabrielseco, line 1: I'm not sure about this lock, let me know if we need to discard it

lukad: This shouldn't be commited. We use [pnpm-lock.yaml](https://github.com/remoteoss/json-schema-form/blob/main/next/pnpm-lock.yaml).

gabrielseco: yes one moment because I was doing huddle with David and now in my "app" is working.

I'll come to see if the PR is needed if not I'll close

antoniocapelo (approved):

PR #188, opened 2025-05-20
fix(next): update required property on fields when it is changed through conditional schemas

lukad:
This fixes a bug in next where the `required` property on a field was not updated when the field's required status changes through a conditional schema.
I've added a test that illustrates the issue.

### DIFF
--- a/next/test/fields/mutation.test.ts
+++ b/next/test/fields/mutation.test.ts
@@ -350,4 +350,97 @@ describe('field mutation', () => {
       })
     })
   })
+
+  it('correctly updates required on fields', () => {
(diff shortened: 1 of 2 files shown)


next/test/fields/mutation.test.ts
```diff
@@ -350,4 +350,97 @@ describe('field mutation', () => {
       })
     })
   })
+
+  it('blah', () => {
```
antoniocapelo, line 354: bleh? :D

lukad: fixed :)

eng-almeida (approved):

PR #189, opened 2025-05-20
fix(next): take x-jsf-presentation options into account when generating options

lukad:


next/src/field/schema.ts
```diff
@@ -332,15 +332,19 @@ export function buildFieldSchema(
     })
   }
 
-  // Handle options
-  const options = getFieldOptions(schema)
-  if (options) {
-    field.options = options
-    if (schema.type === 'array') {
-      field.multiple = true
+  // Generate options from schema if no options were provided via x-jsf-presentation
```
antoniocapelo, line 335: **Suggestion**

Seems to be that all these options-related could be encapsulated in a separate function, similar to what we do with `addCheckboxAttributes`

lukad: We were previously always overriding options from `x-jsf-presentations` which were already spread onto the field in the code block above. Suggestions for a better name are welcome ;)

I'll move this options & fields logic into separate functions as you suggested. That should make this a bit leaner.

antoniocapelo (approved):

PR #191, opened 2025-05-27
fix: check for emptiness of required array/object properties values

antoniocapelo:
When evaluating the `required` errors, we were only checking for the null/undefined values, but arrays/objects might come pre-filled in the form values as empty arrays/objects. This PR addresses it.


eng-almeida (approved):

antoniocapelo:
@joaoalmeida-remote can I get a re-review?

PR #194, opened 2025-06-04
feat(next): add allowForbiddenValues validation option

lukad:
This adds an option that allow passing values to `false` schemas. This would normally result in a validation error but in v0 we explicitly clear values provided to a false schema: https://github.com/remoteoss/json-schema-form/blob/4efb07e3cd27a2e0a79b704e57e12432c80310f2/src/helpers.js#L451-L461

This option will achieve the same goal, but in a more straight forward way. It will also be disabled by default.

### DIFF
--- a/next/src/validation/schema.ts
+++ b/next/src/validation/schema.ts
@@ -21,6 +21,12 @@ export interface ValidationOptions {
    * @default false
    */
   treatNullAsUndefined?: boolean
+
+  /**
+   * If true, providing a value for a schema that is `false`
+   * @default false
+   */
+  allowForbiddenValues?: boolean
 }
 
 /**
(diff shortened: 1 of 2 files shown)


next/src/validation/schema.ts
```diff
@@ -21,6 +21,12 @@ export interface ValidationOptions {
    * @default false
    */
   treatNullAsUndefined?: boolean
+
+  /**
+   * If true, providing a value for a schema that is `false`
```
antoniocapelo, line 26:
```suggestion
   * If true, providing a value for a schema that is `false` won't create an error
```

antoniocapelo (dismissed):

dragidavid (approved):

PR #196, opened 2025-06-09
feat(next): Custom user defined validations

SimoneErba:
Allow users to write their custom functions to use in jsonLogic. We register them once when we create the form. Added `customJsonLogicOps` to the `ValidationOptions` that is a dictionary of name: function

I added the test in the v0 file because the new one doesn't use createHeadlessForm and mocks the jsonLogic


lukad:
Hi @SimoneErba! Thanks for your PR and first of all, sorry for the long delay in getting back to you.

I'd like to accept your PR, but we need to make some adjustments. Unfortunately json-logic has no concept of scopes, instances or similar which means it would be possible for multiple uses of createHeadlessForm to add conflicting custom operators. This could happen in a SPA rendering many different forms.

My proposal here is that we only add the custom ops before validation via `add_operation` and then remove them afterwards via `rm_operation`. This should prevent possible conflicts and keep the global json-logic state clean.

SimoneErba:
Hi @lukad, thanks for your review! It makes sense, I changed the code to add and remove the functions everytime we run the validation

antoniocapelo:
Hi @SimoneErba , there's a small issue with one of the lint jobs, if you can take a look we can move this to the finish line 🙂

SimoneErba:
Hi @antoniocapelo, can you run the CI again to see if it's all good now? Thanks

next/src/form.ts
```diff
@@ -260,6 +261,21 @@ export function createHeadlessForm(
...
+      throw new TypeError('validationOptions.customJsonLogicOps must be an object.')
+    }
+
+    for (const [name, func] of Object.entries(customJsonLogicOps)) {
+      if (typeof func !== 'function') {
+        throw new TypeError(
+          `Custom JSON Logic operator '${name}' must be a function, but received type '${typeof func}'.`,
+        )
+      }
+    }
```
antoniocapelo, line 277: **Suggestion**

Moving this block to a `validateCustomJsonLogicOps` function would make things easier to read 🙂

next/src/form.ts
```diff
@@ -273,18 +289,34 @@ export function createHeadlessForm(
...
   const handleValidation = (value: SchemaValue) => {
-    const updatedSchema = calculateFinalSchema({
-      schema,
-      values: value,
-      options: options.validationOptions,
-    })
+    const customJsonLogicOps = options?.validationOptions?.customJsonLogicOps
+
+    try {
+      if (customJsonLogicOps) {
```
antoniocapelo, line 295: **Suggestion**

Similar approach in the other comment, we could have 2 functions: 

```ts
function addCustomJsonLogicOperations(ops) {
  if (ops) {
          for (const [name, func] of Object.entries(customJsonLogicOps)) {
          jsonLogic.add_operation(name, func)
        }
  }
}

// similar one for removing

```

So this handleValidation would look like:

```ts
const customJsonLogicOps = options?.validationOptions?.customJsonLogicOps

    try {
     addJsonLogicCustomOperations(customJsonLogicOps)

      const updatedSchema = calculateFinalSchema({
        schema,
        values: value,
        options: options.validationOptions,
      })

      const result = validate(value, updatedSchema, options.validationOptions)

      updateFieldProperties(fields, updatedSchema, schema)

      return result
    }
    finally {
     removeJsonLogicCustomOperations(customJsonLogicOps)
    }
```

antoniocapelo:
@SimoneErba I'm sorry about the delay — I'm about to approve the MR but it seems that because of a v1-release file movement, you'll need to also move your new test to the correct path. Sorry for the disturbance — let me know once it's done and I'll approve this + create a new release for the new feature

SimoneErba:
Hi @antoniocapelo, sorry for the delay. In this branch I did a wrong rebase and so I made a new branch from main, but the problem is that if I run `npm run lint` I get many format errors, and if I run `npm run lint -- --fix` I get many changes in the ./json-schema-test-suite folder. Is that fine?  Or am I missing something?

PR #198, opened 2025-06-30
fix(v1): don't fail validation when oneOf array is empty

lukad:
As it turns out, `{"oneOf": []}` should match any value. Previously we would just return a validation error in this case, but I think that's not correct. I think [the spec](https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.1.3) can be interpreted either way but matching any value seems more practical to me and this is also how v0 handles it.


antoniocapelo (approved):
Simple change with tests, looks good to me 👍

PR #199, opened 2025-07-01
chore(next): throw error when deprecated modifyConfig option is passed to createHeadlessForm

lukad:
When migrating code from v0 to v1 this should make it easier to notice when the deprecated `modifyConfig` option is used

### DIFF
--- a/next/src/form.ts
+++ b/next/src/form.ts
@@ -242,10 +242,22 @@ function buildFields(params: { schema: JsfObjectSchema, originalSchema: JsfObjec
   return fields
 }
 
+/**
+ * Ensures that no forbidden options are given
+ * @param options - The options to validate
+ * @throws An error if any forbidden options are found
+ */
+function validateOptions(options: CreateHeadlessFormOptions) {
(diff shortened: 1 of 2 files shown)


next/src/form.ts
```diff
@@ -242,10 +242,22 @@ function buildFields(params: { schema: JsfObjectSchema, originalSchema: JsfObjec
...
 }
 
+/**
+ * Ensures that no forbidden options are given
+ * @param options - The options to validate
+ * @throws An error if any forbidden options are found
+ */
+function validateOptions(options: CreateHeadlessFormOptions) {
+  if (Object.prototype.hasOwnProperty.call(options, 'modifyConfig')) {
+    throw new Error('modifyConfig has been removed from json-schema-form v1')
```
antoniocapelo, line 252:
```suggestion
    throw new Error('`modifyConfig` is a deprecated option and it\'s not supported on json-schema-form v1')
```

antoniocapelo (approved):

antoniocapelo (approved):

PR #211, opened 2025-07-15
chore: Explain ValidationOptions are a backward-compatibility with v0 mistakes

sandrina-p:
It was not well explained what v1 `ValidationOptions` are for, and why they were created. Now it is more clear.

### DIFF
--- a/src/validation/schema.ts
+++ b/src/validation/schema.ts
@@ -18,7 +18,7 @@ export interface ValidationOptions {
    * A null value will be treated as undefined.
    * When true, providing a value to a schema that is `false`,
    * the validation will succeed instead of returning a type error.
-   * This was a bug in v0, we fixed it in v1. If you need the same behavior, set this to true.
+   * This was a bug in v0, we fixed it in v1. If you need the same wrong behavior, set this to true.
    * @default false
    * @example
    * ```ts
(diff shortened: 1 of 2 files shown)


src/validation/schema.ts
```diff
@@ -16,15 +16,28 @@ import { isObjectValue } from './util'
 export interface ValidationOptions {
```
sandrina-p, line 16: note: Now looking back I'd have called this `v0Support` instead of `ValidationOptions`. This is where we'd also pass any config to still support deprecated features. 

Another one that comes to my mind is [dropping `type` attr in favor of of `inputType`](https://linear.app/remote/issue/DEVXP-2862/deprecate-type-attr-on-field). If we have done that already, then it would look like this:

```js
const form = createHeadlessForm(schema, {
  v0Support: {
      includeTypeAttr: true,
  }
}
```

antoniocapelo: indeed 👍

antoniocapelo (approved):

src/validation/schema.ts
sandrina-p: https://github.com/remoteoss/json-schema-form/pull/212/files

PR #212, opened 2025-07-15
fix: rename validationOptions to legacyOptions

sandrina-p:
Follow-up of https://github.com/remoteoss/json-schema-form/pull/211, comment:

> note: Now looking back I'd have called this v0Support instead of ValidationOptions. This is where we'd also pass any config to still support deprecated features.
> 
> Another one that comes to my mind is [dropping type attr in favor of of inputType](https://linear.app/remote/issue/DEVXP-2862/deprecate-type-attr-on-field). If we have done that already, then it would look like this:
> ```js
> const form = createHeadlessForm(schema, {
>   v0Support: {
>       includeTypeAttr: true,
>   }
> }
> ```

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -40,7 +40,7 @@ v1 removes several heavy dependencies:
 +   Field,
 +   FieldType,
 +   FormErrors,
-+   ValidationOptions,
++   V0Support,
 +   ValidationResult
 + } from '@remoteoss/json-schema-form'
 ```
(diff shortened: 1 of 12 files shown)


antoniocapelo:
@sandrina-p I'm a bit uncertain about the `v0Support` name… let's imagine we want to keep this option on a future version. Imagine we introduce a breaking change next week, and we'll be on v2 — having `v0Support` in it will look off.

Can we find a better name? Maybe `Options` + the new JSDoc comments you added mentioning v0 is enough?

PR #213, opened 2025-07-15
fix: properly handle function as custom field properties

antoniocapelo:
We had a bug where custom field properties that are not serializable (eg functions) were being left out in the final output. This PR fixes it.

### DIFF
--- a/src/field/schema.ts
+++ b/src/field/schema.ts
@@ -373,7 +373,12 @@ export function buildFieldSchema({
     return null
   }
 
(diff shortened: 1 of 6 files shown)


sandrina-p (approved):

antoniocapelo:
@sandrina-p I had to fix 2 tests — can you re-approve? 🙏

dragidavid (approved):
Approving this since it seems that Sandrina's approval was reset 👀

PR #214, opened 2025-07-16
chore: improve release scripts

antoniocapelo:
Improve release scripts to remove duplicated prefixes, add check for bump types and generating changelog for official releases

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -132,12 +132,11 @@ async function updateChangelog() {
 
 async function gitCommit({ newVersion, releaseType }) {
   console.log('Committing published version...');
-  const prefix = `v1-${releaseType}`;
 
   let cmd;
   if (releaseType === 'beta' || releaseType === 'official') {
-    // For beta, we commit package.json changes and changelog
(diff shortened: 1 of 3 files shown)


scripts/release.js
```diff
@@ -93,12 +98,18 @@ async function getNewVersion() {
...
+      process.exit(1);
+    }
+
     // get latest official version from git tags
     const tags = await getVersionsFromGitTags();
     const latestOfficialTag = tags.find(tag => !tag.includes('-beta.') && !tag.includes('-dev.'));
     // If no official version found, use v1.0.0 as the starting point
-    const latestOfficialVersion = latestOfficialTag ? latestOfficialTag.replace('v', '') : '1.0.0';
-    return semver.inc(latestOfficialVersion, 'prerelease', 'beta');
+    const latestOfficialVersion = latestOfficialTag ? latestOfficialTag.replace('v', '') : '0.0.0';
```
dragidavid, line 110: Q: Are we resetting this here to 0.0.0?

dragidavid (commented):
Just 1 comment! 👌

scripts/release.js
antoniocapelo: ah, for the **first** release (which already happened), we should consider 0.0.0 as the **last official release** so that the next version came out as 1.0.0

dragidavid (approved):

PR #215, opened 2025-07-17
fix: allow for computed values to be used inside conditionals

antoniocapelo:
### Problem
Computed values from JSON Logic (`x-jsf-logic-computedAttrs`) were not being properly applied when used inside conditional schema structures like `if`/`then`/`else` statements and `allOf`/`anyOf` compositions. This prevented dynamic schema modifications based on computed values within these conditional contexts.

### Solution
- **Extended computed value processing**: Updated `cycleThroughPropertiesAndApplyValues` to recursively process computed values within `if`, `allOf`, and `anyOf` schema structures
- **Re-applied schema rules**: Added logic to re-apply schema rules after computed values are applied, ensuring all conditional logic is properly re-evaluated with the updated schema

### Changes
- **`src/validation/json-logic.ts`**: Extended the computed values processing to handle conditional schemas (`if`, `allOf`, `anyOf`)
- **`src/mutations.ts`**: Added re-application of schema rules after computed values are applied to ensure proper field updates
- **`src/types.ts`**: Updated comment style for consistency
- **`test/validation/json-logic.test.ts`**: Added test case demonstrating computed values working within conditional statements


src/types.ts
```diff
@@ -68,16 +68,16 @@ export type JsfSchema = JSONSchema & {
   // schema (like an if inside another schema), the required property won't be
   // present in the type
   'required'?: string[]
-  // Defines the order of the fields in the form.
+  /* Defines the order of the fields in the form. */
```
dragidavid, line 71: Nit: If we're changing this to be `/*`, we should maybe change the others as well? Especially the multi line comments. 😅

src/validation/json-logic.ts
dragidavid, line 1: Question: would we also need to check for `oneOf` here? 🤔

src/validation/json-logic.ts
```diff
@@ -175,6 +175,27 @@ function cycleThroughPropertiesAndApplyValues(schemaCopy: JsfObjectSchema, compu
   else {
     processProperty(schemaCopy)
   }
+
+  // If the schema has an if statement, we need to cycle through the properties and apply the computed values
+  if (schemaCopy.if) {
+    cycleThroughPropertiesAndApplyValues(schemaCopy.if as JsfObjectSchema, computedValues)
```
dragidavid, line 181: question: just about the casting here. Would that be ok at all times? Would we ever get a different shape?

dragidavid (commented):
Left a comment or 2, otherwise its pretty solid imo!

src/validation/json-logic.ts
antoniocapelo: good point, I've pushed a commit simplifying the types and removing the casts 👍

dragidavid (approved):
Thanks for addressing the comments! All good 👌

PR #223, opened 2025-08-04
feat(next): Custom user defined validations

SimoneErba:
Allow users to write their custom functions to use in jsonLogic. We register them once when we create the form. Added customJsonLogicOps to the oprions that is a dictionary of name: function

I added the test in the v0 file because the new one doesn't use createHeadlessForm and mocks the jsonLogic

I created this new MR becuase in the  [old one](https://github.com/remoteoss/json-schema-form/pull/196) I had problems with the v1 release


lukad:
Hey @SimoneErba, thanks for the MR, this looks great! I'll go ahead and merge this and create a new release.
(not sure why the tests are failing, on main we had an error in `.gitmodules` which I fixed before updating your MR against main... the MR's tests passed on my machine. Must be a github issue).

PR #225, opened 2025-08-18
feat: allow accumulator/current in reduce/map operations

cammellos:
As part of removing some code from the dragon frontend, to be able to use it in our SDK/API, I had to move some frontend logic to the backend.
To do that, I need to use `reduce`, in order to calculate the length on a array. This is in theory supported by `json-logic-js`.
The issue is that we run validation against the data provided, but we don't accommodate for the fact that if we are within a reduce/map block, we should allow `current` and `accumulator`. 
This code enriches the validation data if we are within a reduce/map block, while it maintains checking the integrity of the rest of the variables.

Let me know if the implementation is what you expect or you'd rather be different, thank you!

### DIFF
--- a/v0/src/jsonLogic.js
+++ b/v0/src/jsonLogic.js
@@ -404,26 +404,40 @@ function validateInlineRules(jsonSchema, sampleEmptyObject) {
  * @param {Function} errorMessage - Function to generate custom error message.
  *                                  Receives the invalid rule part and should throw an error message string.
  */
+
 function checkRuleIntegrity(
   rule,
   id,
   data,
   errorMessage = (item) =>
-    `[json-schema-form] json-logic error: rule "${id}" has no variable "${item.var}".`
+    `[json-schema-form] json-logic error: rule "${id}" has no variable "${item.var}".`,
+  inReduceOrMap = false
 ) {
   Object.entries(rule ?? {}).map(([operator, subRule]) => {
     if (!Array.isArray(subRule) && subRule !== null && subRule !== undefined) return;
     throwIfUnknownOperator(operator, subRule, id);
 
+    // If we are within a reduce or map, we allow references to `accumulator` and/or `current`.
+    // We augment the data so that we still validate any other field.
+    const isReduceOrMap = inReduceOrMap || operator === 'reduce' || operator === 'map';
+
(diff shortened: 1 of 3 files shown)


v0/src/tests/jsonLogic.fixtures.js
```diff
@@ -1073,7 +1073,7 @@ export const schemaWithReduceAccumulator = {
           '*': [
             { var: 'working_hours_per_day' },
             {
-              reduce: [{ var: 'work_days' }, { '+': [{ var: ['accumulator', 0] }, 1] }, 0],
+              reduce: [{ var: 'work_days' }, { '+': [{ var: "accumulator" }, 1] }, 0],
```
eng-almeida, line 1076: minor: for consistency we should keep the single quotes.

v0/src/jsonLogic.js
```diff
@@ -404,26 +404,40 @@ function validateInlineRules(jsonSchema, sampleEmptyObject) {
...
     throwIfUnknownOperator(operator, subRule, id);
 
+    // If we are within a reduce or map, we allow references to `accumulator` and/or `current`.
+    // We augment the data so that we still validate any other field.
+    const isReduceOrMap = inReduceOrMap || operator === 'reduce' || operator === 'map';
+
+    const validationData = isReduceOrMap
+      ? {
+          ...data,
+          accumulator: 0,
```
eng-almeida, line 427: Setting the initial value of the accumulator to 0, we're assuming that the final result of the reduce will always be a number. Is this intentional? I'm thinking about a different scenario where you might want to concatenate strings:

```
{"reduce":[
    {"var":"integers"},
    {"+":[{"var":"current"}, {"var":"accumulator"}]},
    ""
]}
```
When running `jsonLogic.apply`, this would result in `0Hello World`, right?

cammellos: mmh, yes, that could be a problem, the logic would work since this is only for checking the integrity of the rules, and when actually running it for your form it will use the right parameters, but it would be a problem if you say have a division instead, let me create a test to see, good point

eng-almeida: Makes sense, thanks for adding the test 👍

PR #226, opened 2025-08-22
feat(json-logic): apply computed attrs on nested allOf/anyOf/oneOf pr…

eshiota:
Refactors the `cycleThroughAttrsAndApplyValues` implementation on `json-logic` with a more recursive-ish logic so that properties nested on `allOf`/`anyOf`/`oneOf` definitions are also parsed when `x-jsf-logic-computedAttrs` are resolved.

For instance, given the following property:

```
'temperature_setting': {
  'title': 'Select a preset temperature',
  'type': 'string',
  'oneOf': [
    {
      'title': 'Low',
      'const': 18,
    },
    {
      'title': 'Medium',
      'const': 20,
    },
    {
      'title': 'Maximum',
      'x-jsf-logic-computedAttrs': {
        'const': 'maximum_temperature',
      },
    },
  ],
},
```

the `const` on the last item of the property's `oneOf` will be properly parsed with the computed value.

### DIFF
--- a/v0/src/jsonLogic.js
+++ b/v0/src/jsonLogic.js
@@ -404,26 +404,40 @@ function validateInlineRules(jsonSchema, sampleEmptyObject) {
  * @param {Function} errorMessage - Function to generate custom error message.
  *                                  Receives the invalid rule part and should throw an error message string.
  */
+
 function checkRuleIntegrity(
   rule,
   id,
   data,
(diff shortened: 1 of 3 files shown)


dragidavid (approved):
Forgot to approve last night 🤦 My bad! 

Looks good

PR #231, opened 2025-09-15
fix(array): Fix conditional visibility in group-arrays for booleans conditionals

gerardabello:
Group-arrays conditionals that affect visibility are not working
As the comment explains, it does not fix all the logic in group-array, but it fixes some of it (coincidentally the logic I need 😄)


src/mutations.ts
```diff
@@ -133,7 +133,25 @@ function applySchemaRules(
...
+             *
+             * This means that the visibility rules for group-array items will only be based on the
+             * schema and will not work based on the actual values of the items in the array.
+             *
+             * This is not ideal, but it's better than the previous situation where the rules were not applied at all.
+             */
+            applySchemaRules(propertySchema.items as JsfObjectSchema, {}, options, jsonLogicContext)
+          }
+          else {
+            applySchemaRules(propertySchema.items as JsfObjectSchema, values[key], options, jsonLogicContext)
```
sandrina-p, line 153: Why do we need a if/else here? AFAIK every field with `items` is a group-array and the value it's an array too.

I also tried to run all the tests with this else commented out - and tests still passed. 

I'd appreciate @lukad eyes and expertise here to confirm my assumption

sandrina-p: So, passing a value as `{}` works because `applySchemaRules()` does a early returns if the value is not an array (which would always return falsy, even if the condition matches). 
As the `if` is hardcoded (a boolean true/false), the `validateSchema` doesn't even evaluate the value itself, regardless if it's `{}` or something else. It only looks at `typeof schema === 'boolean'`.

I believe the root bug is somewhere at `validateArray()`... still checking

sandrina-p: Which is basically what @gerardabello wrote in the comments, but it's hard to grasp until you really go through the code. I need to drop, this is where I'm left. I'd need considerable more time to debug this properly, which I can't right now.

Okay I wanna unblock you @gerardabello. Here's the plan:
1. Remove the redundant conditional
2. Write the extra test I asked
4. Open a MR in Dragon to ensure all tests still pass with a dev ver
5. We merge and release this

Thank you @gerardabello for your patience 🙇

test/fields/visibility.test.ts
```diff
@@ -5,6 +5,144 @@ import { getField } from '../../src/utils'
...
+                    'title': 'Field 2',
+                    'type': 'string',
+                  },
+                },
+                'x-jsf-order': ['full_name', 'field_1', 'field_2'],
+                'required': ['full_name'],
+                'type': 'object',
+                'allOf': [
+                  {
+                    if: true,
```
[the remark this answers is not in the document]
gerardabello: done

sandrina-p (approved):

test/fields/array.test.ts
```diff
@@ -795,6 +795,243 @@ describe('buildFieldArray', () => {
...
+        }
+
+        const form = createHeadlessForm(schema, { initialValues: { children: [{ full_name: 'name' }] } })
+
+        expect(form.fields.find(({ name }) => name === 'children')?.fields?.find(({ name }) => name === 'field_1')?.isVisible).toBe(false)
+        expect(form.fields.find(({ name }) => name === 'children')?.fields?.find(({ name }) => name === 'field_2')?.isVisible).toBe(true)
+      })
+    })
+
+    describe.skip('with logic based on answers', () => {
```
sandrina-p, line 937: I've created an [internal ticket](https://linear.app/remote/issue/RMT-1745/jsf-grouparray-conditional-fields-dont-work-visibility) for this

PR #233, opened 2025-09-25
fix(options): spread all option's presentation properties into option's root

eshiota:
Spreads all `x-jsf-presentation` properties inside a radio option's schema into the option's final representation, instead of just picking the `meta` key.

This maintains the `v1` feature parity by implementing the same behaviour as `v0` (https://github.com/remoteoss/json-schema-form/pull/17)

### DIFF
--- a/src/field/schema.ts
+++ b/src/field/schema.ts
@@ -160,8 +160,7 @@ function convertToOptions(nodeOptions: JsfSchema[]): Array<FieldOption> {
     .map((schemaOption) => {
       const title = schemaOption.title
       const value = schemaOption.const
-      const presentation = schemaOption['x-jsf-presentation']
-      const meta = presentation?.meta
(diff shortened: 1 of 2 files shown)


src/field/schema.ts
```diff
@@ -172,15 +171,10 @@ function convertToOptions(nodeOptions: JsfSchema[]): Array<FieldOption> {
         value,
       }
 
-      // Add meta if it exists
-      if (meta) {
-        result.meta = meta
-      }
-
       // Add other properties, without known ones we already handled above
       const { title: _, const: __, 'x-jsf-presentation': ___, ...rest } = schemaOption
```
danielcardoso5, line 175: **nit (non-blocking)** We could do this destructuring above since we already grab `title` and `const` from `schemaOptions`

```
const { title, const: value, 'x-jsf-presentation': jsfPresentation, ...rest } = schemaOption;
const presentation = typeof jsfPresentation === 'object' ? jsfPresentation : {};
```

danielcardoso5 (approved):

PR #237, opened 2025-10-17
fix(json-logic): add custom operators on headless form creation

eshiota:
Currently all custom JSON Logic operators are applied ad-hoc when the validation functions are executed, and then removed afterwards. While this makes sense from the point of view of simulating a deterministic call to JSON Logic, other places on the form's lifecycle require these custom operators to be available.

This means that if a form is loaded with an initial value where a property needs a computed value that relies on a custom operator, the operator won't be available (as it's only added when a validation happens), thus throwing an error.

This PR simplifies the logic by adding all custom operators at once when the `createHeadlessForm` is called, which happens only once per page load.


danielcardoso5 (changes requested):
@eshiota needs tests fixing

PR #239, opened 2025-10-23
chore(performance): optimise field updates with large dropdown options

ollyd:
Forms containing large select option lists (e.g. country dropdown) caused severe UI lag on every keystroke in other text field inputs. `deepMergeSchemas` was recursively merging hundreds of option objects per validation.

**Changes:**

- Added content-hash caching for field options
- Skip deep merge for options (replace arrays instead)

No more discernible lag, with a measurable ~10x improvement:

Before:
Screenshot 2025-10-23 at 10 23 58 am

After:
Screenshot 2025-10-23 at 10 19 37 am

Related internal [MR](https://gitlab.com/remote-com/employ-starbase/dragon/-/merge_requests/49412).

### DIFF
--- a/src/field/schema.ts
+++ b/src/field/schema.ts
@@ -154,24 +154,7 @@ function hashOptions(opts: JsfSchema[]): string {
     return '0'
   }
 
-  // Extract the const value from an option
-  const extractValue = (option: JsfSchema) => {
-    return (typeof option === 'object' && option !== null) ? option.const : option
-  }
-
-  const length = opts.length
-  const start = opts[0]
-  const middle = opts[Math.floor(length / 2)]
-  const end = opts[length - 1]
-
-  // Sample 3 parts of the array for a reliable hash
-  const sampledValues = [
-    extractValue(start),
-    extractValue(middle),
-    extractValue(end),
-  ]
-
-  return `${length}:${JSON.stringify(sampledValues)}`
+  return JSON.stringify(opts)
 }
 
 /**
--- a/test/fields/options.test.ts
+++ b/test/fields/options.test.ts
@@ -1,0 +1,157 @@
+import { describe, expect, it } from '@jest/globals'
+import { createHeadlessForm } from '../../src/form'
+
+describe('Select field options', () => {
+  it('should return cached options based on content hash', () => {
+    // Create two separate oneOf arrays with identical content
+    const options1 = [
+      { const: 'value_1', title: 'Option 1' },
+      { const: 'value_2', title: 'Option 2' },
+      { const: 'value_3', title: 'Option 3' },
+    ]
+
+    const options2 = [
+      { const: 'value_1', title: 'Option 1' },
+      { const: 'value_2', title: 'Option 2' },
+      { const: 'value_3', title: 'Option 3' },
+    ]
+
+    // Different object references but same content
+    expect(options1).not.toBe(options2)
+    expect(options1).toEqual(options2)
+
+    const schema = {
+      type: 'object' as const,
+      properties: {
+        field1: {
+          type: 'string' as const,
+          oneOf: options1,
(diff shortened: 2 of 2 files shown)


src/field/schema.ts
```diff
@@ -142,6 +142,38 @@ You can fix the json schema or skip this error by calling createHeadlessForm(sch
...
+  const extractValue = (option: JsfSchema) => {
+    return (typeof option === 'object' && option !== null) ? option.const : option
+  }
+
+  const length = opts.length
+  const start = opts[0]
+  const middle = opts[Math.floor(length / 2)]
+  const end = opts[length - 1]
+
+  // Sample 3 parts of the array for a reliable hash
```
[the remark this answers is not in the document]
sandrina-p: suggestion: Perhaps to be more sure that 2 or 3 more options from the middle? Like 1/3, 1/2, and 2/3?

src/field/schema.ts
```diff
@@ -142,6 +142,38 @@ You can fix the json schema or skip this error by calling createHeadlessForm(sch
...
+  const sampledValues = [
+    extractValue(start),
+    extractValue(middle),
+    extractValue(end),
+  ]
+
+  return `${length}:${JSON.stringify(sampledValues)}`
```
sandrina-p: why do we need an array and then stringify? Couldn't we simplify to just a string:

```js

const sampledValues = `${extractValue(start)},${extractValue(middle)},${extractValue(end)}`

return `${length}:${sampledValues}`
```

src/utils.ts
```diff
@@ -99,6 +99,14 @@ export function deepMergeSchemas<T extends Record<string, any>>(schema1: T, sche
...
     else if (schema1Value && Array.isArray(schema2Value)) {
       const originalArray = schema1Value
+
+      // For 'options' arrays, just replace the whole array (they're immutable and cached) rather
+      // than recursively deep merging them
+      if (key === 'options') {
+        schema1[key as keyof T] = schema2Value as T[keyof T]
+        continue
+      }
+
```
sandrina-p: I tested locally, it works nice, well done @ollyd !!

Consider this approved ☺️

src/field/schema.ts
```diff
@@ -153,7 +185,13 @@ You can fix the json schema or skip this error by calling createHeadlessForm(sch
  * If it doesn't, we skip the option.
  */
 function convertToOptions(nodeOptions: JsfSchema[]): Array<FieldOption> {
-  return nodeOptions
+  const hash = hashOptions(nodeOptions)
+  const cached = optionsMap.get(hash)
+  if (cached) {
```
sandrina-p: Have you considered any type of unit tests here? 👀 

I understand it might be trickier because of implementation details and lifecycle. So at very least we can leave a code comment saying there's no unit tests. To carefully test manually when changing the logic.

sandrina-p (dismissed):

src/field/schema.ts
ollyd: Yeah I think you're right, thanks. I had it that way at first, but then I thought of sampling the array because hashing the entire options could be massive. In reality it's a micro optimisation that wouldn't be noticeable to a user though. Better to rule out any potential edge cases. I'll change that.

src/field/schema.ts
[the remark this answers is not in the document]
ollyd: Yes it was a very basic hash implementation as I didn't want to introduce any deps, and I thought it would be ok but saying that, if an edge case is possible, it will happen so I will update it! 🙂

src/utils.ts
ollyd: Hey @sandrina-p, I've added some tests and updated the hashing to stringify the entire options. If you're happy with the new changes I'll get this merged in!

PR #258, opened 2026-05-26
fix(schema): allow empty array for required field with minItems

spiderocious:
Fixes a bug where a `required` array property fails validation when given an
empty array, even though the schema explicitly allows it via `minItems: 0`.
A required `group-array` with `min: 0` could never be submitted empty.

## Changes Made

- **`src/validation/schema.ts`**
  - In the `required` check, an empty array is no longer automatically treated
    as a missing value when the field schema declares `minItems`. In that case
    the array counts as *present*, and its length is validated by `minItems`
    (which runs separately) rather than by `required`.
  - Behaviour is unchanged for required arrays that do **not** declare
    `minItems`: an empty array is still treated as missing, preserving existing
    behaviour.

- **`test/validation/required_field.test.ts`**
  - Added a test that a required array with `minItems: 0` accepts an empty array.
  - Added a test that a required array with `minItems: 1` reports a `minItems`
    error (not a `required` error) for an empty array, confirming length is
    still enforced by the right mechanism.
  - Added a test that a required array without `minItems` still treats an empty
    array as missing.

## Related Issues/PRs
Fixes #247

## Checklist

- [x] I have read the [Contributing Guidelines](../../CONTRIBUTING.md) and followed them.
- [x] Unit tests were added.

---

> [!NOTE]
> **Medium Risk**
> Changes core required-field semantics for all required arrays (not only minItems: 0), which may alter validation for forms that relied on empty [] triggering "Required field".
> 
> **Overview**
> Fixes validation so **`required` on array properties no longer fails when the value is `[]`**. In `schema.ts`, the required-key filter used to treat empty arrays like missing fields; arrays (including empty) are now always considered present, with comments pointing authors to **`minItems`** for non-empty requirements.
> 
> **Tests** were updated: a required `list: []` case in array field tests now expects no error, and `required_field.test.ts` adds coverage for empty vs populated arrays, **`minItems: 0` / `1`**, and required arrays without **`minItems`** (empty array valid; missing key still **Required field**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05b9fd4b319b99c58217bd827fb9f00ca7110f05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/validation/schema.ts
+++ b/src/validation/schema.ts
@@ -247,6 +247,15 @@ export function validateSchema(
       // - it's null AND treatNullAsUndefined option is true
       // - it's an array/object and it's empty
       if (Array.isArray(fieldValue)) {
+        // By default an empty array is treated as a missing required value.
+        // However, when the field schema explicitly declares `minItems`, the
+        // allowed length is governed by `minItems` (validated separately), so
+        // an empty array should be considered present. This lets a required
+        // array opt into accepting empty values via `minItems: 0`.
+        const fieldSchema = schema.properties?.[key]
+        if (fieldSchema && typeof fieldSchema !== 'boolean' && fieldSchema.minItems !== undefined) {
+          return false
+        }
         return fieldValue.length === 0
       }
 
--- a/src/validation/schema.ts
+++ b/src/validation/schema.ts
@@ -245,18 +245,15 @@ export function validateSchema(
       // Field is considered missing if:
       // - it's undefined OR
       // - it's null AND treatNullAsUndefined option is true
-      // - it's an array/object and it's empty
+      // - it's an object and it's empty
+      //
+      // An array (including an empty one) counts as a present value: `required`
+      // only checks for the presence of a key, while array length is governed
+      // by `minItems` / `maxItems` (validated separately). To require a
+      // non-empty array, use `minItems: 1` rather than relying on `required`.
+
       if (Array.isArray(fieldValue)) {
-        // By default an empty array is treated as a missing required value.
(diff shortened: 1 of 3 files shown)


spiderocious:
@sandrina-p  @danielcardoso5 please review

src/validation/schema.ts
```diff
@@ -247,6 +247,15 @@ export function validateSchema(
       // - it's null AND treatNullAsUndefined option is true
       // - it's an array/object and it's empty
       if (Array.isArray(fieldValue)) {
+        // By default an empty array is treated as a missing required value.
```
sandrina-p: bug: This is wrong (an existing bug). By default an empty array should NOT be treated as a missing required value. I confirmed in the jsonschemalint.com

Simple JSON Schema:

```json
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "browser": {
      "title": "Browsers",
      "type": "array"
    }
  },
  "required": [
    "browser"
  ]
}
```

Image

So we should adapt the code, tests and comments for both scenarios.

spiderocious: Thanks for the review feedback, @sandrina-p , I'll address this by simplifying the check so an empty array always counts as present, and I'll update the comments and tests (including the pre-existing one in array.test.ts) to match. Pushing the fix shortly.

spiderocious: I've made the changes, please check again.  @sandrina-p

PR #261, opened 2026-06-22
fix(v0): validation was crashing when a conditional reaches into a undefined fieldset value

sandrina-p:
## What

V0 crashes when a schema conditional reaches into a nested fieldset whose value is `null`/`undefined`. This happens for rows/payloads that omit a fieldset, throwing instead of validating.

## Example

Schema with a conditional that reaches into the `pet` fieldset:

```json
{
  "type": "object",
  "properties": {
    "pet": {
      "type": "object",
      "x-jsf-presentation": { "inputType": "fieldset" },
      "properties": {
        "species": { "type": "string", "enum": ["dog", "cat"] }
      }
    },
    "pet_name": { "type": "string" }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "pet": { "properties": { "species": { "const": "dog" } } }
        },
        "required": ["pet"]
      },
      "then": { "required": ["pet_name"] }
    }
  ]
}
```

Validating a value that omits the fieldset crashes:

```js
form.handleValidation({ pet: null }) // TypeError: Cannot read properties of null (reading 'species')
```

## Root cause

`checkIfConditionMatchesProperties` recurses into the fieldset with `formValues["pet"]` which has a null value, so then it crashes inside.

## Fix

Coalesce to `{}` (`formValues[name] ?? {}`) so the nested condition evaluates instead of crashing.

REF CPF-3408.

🤖 co-written with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/v0/package.json
+++ b/v0/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.12.3-beta.0",
+  "version": "0.12.4-dev.20260622162258",
   "description": "Headless UI form powered by JSON Schemas",
   "author": "Remote.com <engineering@remote.com> (https://remote.com/)",
   "license": "MIT",
--- a/v0/package.json
+++ b/v0/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.12.4-dev.20260622162258",
+  "version": "0.12.3-beta.0",
   "description": "Headless UI form powered by JSON Schemas",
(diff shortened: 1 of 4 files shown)


v0/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remoteoss/json-schema-form",
-  "version": "0.12.3-beta.0",
+  "version": "0.12.4-dev.20260622162258",
```
sandrina-p: New version to test in our internal repos:  `0.12.4-dev.20260622162258`.

Reverting it, so we can merge this after internal repos CI is green

HanliTheron (approved):

PR #193, opened 2025-05-29
Fix/fix some backward compatible code

antoniocapelo:

PR #263, opened 2026-07-03
feat: replace option-like arrays via conditionals

danielcardoso5:
Matches v0 behavior by allowing to override option-like arrays present in `oneOf` / `anyOf` properties. This allows to change the list of options allowed for fields based on conditionals.

Example:

<details><summary>Schema</summary>
<p>

```json
{
    "type": "object",
    "properties": {
        "trigger": {
            "type": "string"
        },
        "choice": {
            "type": "string",
            "oneOf": [
                {
                    "const": "a",
                    "title": "A"
                },
                {
                    "const": "b",
                    "title": "B"
                },
                {
                    "const": "c",
                    "title": "C"
                }
            ]
        }
    },
    "allOf": [
        {
            "if": {
                "properties": {
                    "trigger": {
                        "const": "only c"
                    }
                },
                "required": [
                    "trigger"
                ]
            },
            "then": {
                "properties": {
                    "choice": {
                        "oneOf": [
                            {
                                "const": "c",
                                "title": "C"
                            }
                        ]
                    }
                }
            }
        }
    ]
}
``` 

</p>
</details> 

| Before | After |
|--------|--------|
| image |  image|

---

> [!NOTE]
> **Medium Risk**
> Changes core schema merge logic used when applying conditionals and updating fields; wrong array handling could break validation or option lists across many forms.
> 
> **Overview**
> **`deepMergeSchemas`** now **fully replaces** schema arrays when a conditional branch supplies new values—except **`required`**, which still only **appends** missing entries. That replaces the old behavior that special-cased `options` and tried to merge array contents (including a buggy recursive merge on object items).
> 
> Conditional **`allOf` / `if` / `then`** merges can therefore **narrow or swap** field choices: **`oneOf`**, **`anyOf`**, **`enum`**, and presentation **`options`** update on validation and **revert** when the branch is inactive, with validation aligned to the active set.
> 
> Tests cover **`deepMergeSchemas`** edge cases, **`convertKBToMB`**, and headless-form scenarios for conditional **`oneOf`**, **`anyOf`**, and **`enum`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f06bb2adbefa1fa8ba974e06b39f9d828f1fcf84. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danielcardoso5:
[Internal pipeline](https://gitlab.com/remote-com/employ-starbase/dragon/-/pipelines/2656134842) is 🟢

test/fields/options.test.ts
```diff
@@ -155,3 +155,207 @@ describe('Select field options', () => {
     expect(field2Options?.[2]).toEqual({ label: 'Option D', value: 'value_d' })
   })
 })
+
+describe('conditionally replacing option-like arrays', () => {
+  const getOptions = (form: ReturnType<typeof createHeadlessForm>, name: string) =>
+    form.fields.find(f => f.name === name)?.options
+
+  it('should fully replace oneOf options when a conditional branch provides new options', () => {
```
sandrina-p: does this matches the official specs? https://jsonschemalint.com/#!/version/2020-12/markup/json

I have the impression this is not a bug in v1, it's a feature. Sorry I didn't test, i'm dropping off right now

src/utils.ts
danielcardoso5: The code here could be much simpler if we don't do this deep merging of array by index. I'm not sure if this is intended behavior as I could not find a single test that breaks when this is removed (besides the one I added). It's not clear to me the purpose here.

We're also doing a redundant `deepMergeSchemas` for every item in `schema2Value`, the call never changes. We could simply do it once if one of the items in `schema2Value` is an object.

Thoughts @sandrina-p @eshiota? Do you know of a case where we need positional merging of array items?

test/fields/options.test.ts
danielcardoso5: ah, need to update the tests here. A conditional branch should not introduce new options, but should be able to override existing ones.

Example:
<details><summary>schema</summary>
<p>

```json
{
    "type": "object",
    "properties": {
        "trigger": {
            "type": "string"
        },
        "choice": {
            "type": "string",
            "oneOf": [
                {
                    "const": "a",
                    "title": "A"
                },
                {
                    "const": "b",
                    "title": "B"
                },
                {
                    "const": "c",
                    "title": "C"
                }
            ]
        }
    },
    "allOf": [
        {
            "if": {
                "properties": {
                    "trigger": {
                        "const": "only c"
                    }
                },
                "required": [
                    "trigger"
                ]
            },
            "then": {
                "properties": {
                    "choice": {
                        "oneOf": [
                            {
                                "const": "c",
                                "title": "C"
                            }
                        ]
                    }
                }
            }
        }
    ]
}
```

</p>
</details> 

Testing in https://jsonschemalint.com/#!/version/2020-12/markup/json with:

❌
 ```json
{
  "trigger": "only x",
  "choice": "b"
}
``` 

✅ 
```json
{
  "trigger": "only x",
  "choice": "x"
}
```

So it stops accepting some of the base options, which means they've been replaced by the conditional branch.

danielcardoso5: What our code doesn't do is validating if the option being included in a conditional branch is present in the default options config of the field. Per the spec, an option introduced in a conditional branch is never valid. We can add that validation though to prevent new options from being introduced.

sandrina-p: yes, exactly, that it is. I didn't remember it exactly. thanks!

src/utils.ts
sandrina-p: I don't remember exactly why, and it sucks tests weren't added specific for this. It seems it came from this PR https://github.com/remoteoss/json-schema-form/pull/193. Asking Opus also doesn't spark any reason behind it?

It's a risk, but I agree with simplifying it. Then we'll find out why it's needed, and we revert it.

sandrina-p: Ah, i remember a possibility! It's for those cases when we mutate the JSON Schema and add functions `Component: React`, to keep it.

danielcardoso5: @sandrina-p could you provide an example? IMO this would only make sense if we're merging an array with exactly the same elements, but to provide additional keys or override existing ones. This behavior is exactly the reason why the options merging in my PR description example result in "C, B, C" options. It is replacing at position 0 the option with the one provided via the conditional while preserving the remaining ones.

danielcardoso5: > Asking Opus also doesn't spark any reason behind it?

I did this, here's the output:

<details><summary>reasoning</summary>
<p>


## What the index-merge actually is

At line 125–126, when an array item is an object, it calls `deepMergeSchemas(originalArray, schema2Value)`. Since arrays are objects in JS, `Object.entries` yields index keys (`"0"`, `"1"`, …), so this becomes a **positional overlay**: `schema2Value[0]` merges into `originalArray[0]`, `[1]` into `[1]`, etc. It's a left-join by index — only indices present in `schema2Value` are touched; extra base elements survive, and base properties the branch didn't mention are preserved.

## When index-alignment is meaningful (the scenarios where it's needed)

It only makes sense when the two arrays are **positionally-aligned representations of the same ordered set**:

1. **Tuple schemas** — `prefixItems` / array-form `items`. Here index *is* semantics: `prefixItems[0]` describes element 0 of the value. If a branch supplies `prefixItems: [{ minLength: 1 }, {}]`, you want it overlaid onto the base's `[{ title:'First' }, { title:'Second' }]` position-for-position → `[{title:'First', minLength:1}, {title:'Second'}]`. Merging by index is exactly right.

2. **jsf `Field` sub-field arrays** (`fields`) — the dominant real case. A fieldset's base has sub-fields `[name, age]` in a stable order. A conditional branch re-derives the *same* fields in the *same* order but flips, say, `age` to required/hidden. Index-merge overlays that one change while keeping every other base property. This works precisely because the branch is a recomputation of the same list, not an arbitrary new list.

</p>
</details> 

Could be explained by the call of `deepMergeSchemas` via `updateFieldProperties` when dealing with nested fields of a fieldset. I'll give it a few tests to see where it plays a role

eshiota: Honestly I can't think of a valid reason to keep the current behaviour of merging arrays. I might be wrong but even fieldsets should not require it.

I say we change it so that arrays are overridden and publish a new minor version with a beta tag. This will make the overall behaviour easier to understand. Otherwise we'll live with this discussion going on forever 😅

sandrina-p: I've asked Claude, I know i've worked in things related with array merges, but I don't recall what exactly, it has been +2 years. But Claude found it! TLDR: Yes, let's simplify it!

> **Background on the array index-merge in `deepMergeSchemas`**
>
> No confirmed intentional reason for it. It's the same category of bug I (Sandrina) have already fixed in `modify()` (PR #82) — where arrays were being merged by index instead of replaced, causing leftover values from the base schema to bleed through. That fix added a `mergeReplaceArray` customizer to lodash's `mergeWith`. `deepMergeSchemas` in `utils.ts` is a separate hand-rolled function that never got the same fix. There's no test that specifically covers or validates the index-merge behavior, and I don't recall adding it intentionally. Overriding arrays is the right call.

Check the PR #82, the scenario that needs to be tested here too (which you added), so we are good to go with simplifying stuff

danielcardoso5: Yep, I noticed that the `modify` code avoids this and just replaces the whole array. I do have some tests hitting the `deepMergeSchemas` of arrays logic but they don't break when removing it. Will check those ones to see if we're missing something

danielcardoso5: > Could be explained by the call of deepMergeSchemas via updateFieldProperties when dealing with nested fields of a fieldset. I'll give it a few tests to see where it plays a role

I've confirmed this. The scenarios where deep merge of arrays happened were all via `updateFieldProperties` calls, but from my testing they were not necessary because we send a complete array (same number of elements) that can replace the original one, no need to do positional merging. This logic ran for schemas with fieldsets, it ran for [this one](https://github.com/remoteoss/json-schema-form/blob/main/test/fields/visibility.test.ts#L155) and [this one](https://github.com/remoteoss/json-schema-form/blob/main/test/fields/visibility.test.ts#L258).

I've simplified the code and tested against a complex schema we have with nested fieldset, it's working as expected.

New 🟢 [pipeline with latest changes](https://gitlab.com/remote-com/employ-starbase/dragon/-/pipelines/2661068559).

sandrina-p: thanks! I gave a last test too locally (JSF + Playground) and didn't find anything odd either. Let's go 🚢

sandrina-p (approved):

src/utils.ts
sandrina-p: ah wait, let me do one final test

sandrina-p: I found another bug, but it also happens in the current JSF `main`, so approving anyway. 🚢 

The bug / feature: The conditional options to work must be a _segment_ of the original options, but never an _extension_.
For example. If original options is [A, B], and the conditional options is [A, C]. A works, but C doesn't. Hard to explain in words, check this out:

Schema

```json
{
  "type": "object",
  "properties": {
    "userType": {
      "type": "string"
    },
    "permissions": {
      "enum": ["a", "b"]
    }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "userType": {
            "const": "guest"
          }
        },
        "required": [
          "userType"
        ]
      },
      "then": {
        "properties": {
          "permissions": {
            "enum": ["a", "c"]
          }
        }
      }
    }
  ]
}
````

Data payloads:

```jsonc
// Valid
{
  "userType": "foo",
  "permissions": "a" 
}

{
  "userType": "guest",
  "permissions": "a" 
}

// Not valid! Because C is not part of the original options
{
  "userType": "guest",
  "permissions": "c" 
}
```

image


For it to work, the original field type must be wider (a string, not enum), and all options need to be conditional inside the if/then/else. This applies for both `enum`, `anyOf` and `oneOf`. example:

image

danielcardoso5: @sandrina-p that's the behavior I mentioned in [this comment](https://github.com/remoteoss/json-schema-form/pull/263#discussion_r3530855234).

> We can add that validation though to prevent new options from being introduced.

My proposal would be to do this for v1, so that conditionals cannot introduce new options that are not present in the original options. So more specifically, we can:

1. for option-like arrays, grab all incoming options
2. for each incoming option, check if it's part of the field's default options 
3. if it is add it, if not omit it (we can log a warning when omitting)
4. do the normal replace

I would prefer a separate PR for that since it's a different fix of what I'm doing here. I'll push one for that if you think the solution makes sense.

sandrina-p: yes, yes, separate follow-up MR, not this one. I've already approved this one

PR #264, opened 2026-07-10
fix: modify() preserve unknown top-level schema keys when using pick

sandrina-p:
## Bug

This bug happens in v1, but not in v0.

`modify()` used with `pick` silently drops any top-level schema key it doesn't explicitly know about (`type`, `x-rmt-meta`, `additionalProperties`, `$schema`, etc). Only `properties`, `x-jsf-order`, `required`, `allOf`, `x-jsf-logic` survived.

Repro:

```js
const schema = {
  'type': 'object',
  'x-rmt-meta': { jsfOldVersion: true },
  'properties': {
    name: { type: 'string' },
    age: { type: 'number' },
  },
}

const { schema: result } = modify(schema, { pick: ['name'] })

result.type // undefined, should be 'object'
result['x-rmt-meta'] // undefined, should be { jsfOldVersion: true }
```

## Fix

Added the missing `default` case so unrecognized top-level keys pass through unchanged, matching v0.

### DIFF
--- a/src/modify-schema.ts
+++ b/src/modify-schema.ts
@@ -354,9 +354,8 @@ function pickFields(originalSchema: JsfSchema, fieldsToPick: ModifyConfig['pick'
 
         break
       }
-      case 'x-jsf-logic':
+      default:
         newSchema[attrKey] = attrValue
-        break
     }
   })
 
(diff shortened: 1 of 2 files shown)


nitin-remote (approved):
Looks good, thank you!

PR #267, opened 2026-07-24
fix: add option for new conditional options behavior

danielcardoso5:
Follow up to the breaking change introduced in https://github.com/remoteoss/json-schema-form/pull/265.

It adds a new `disallowNewConditionalOptions`(defaults to `false`) that allows to opt into the breaking change behavior: disallowing new conditional options from being introduced by conditional logic.

This ensures existing usage does not break when https://github.com/remoteoss/json-schema-form/pull/265 is released.

---

> [!NOTE]
> **Medium Risk**
> Changes how conditional schemas affect field options and validation for consumers who opt in or rely on introducing new options via branches; default behavior is preserved but deprecation warnings may surface in production consoles.
> 
> **Overview**
> Adds **`disallowNewConditionalOptions`** to `createHeadlessForm` (default **`false`**) so teams can adopt stricter conditional option merging without a breaking release.
> 
> With the flag **`true`**, `mergeSchemaBranch` only keeps branch **`enum` / `oneOf` / `anyOf` / presentation `options`** that already exist on the base field (narrow or re-label); values the branch adds are dropped. If the base has no option array for that key, branch options can still be applied. With the default **`false`**, branches still **replace** option arrays wholesale (legacy), but a **one-time `console.warn`** points callers at the new flag when a branch would add options not on the base.
> 
> Conditional rule application now passes full **`CreateHeadlessFormOptions`** into `mergeSchemaBranch` via `mutations.ts`. **MIGRATING.md** documents the migration path; tests cover both modes. Unrelated: removes unused **`convertDiskSizeFromTo`** from `utils`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6db7a33eb9b525aefd0eb7a75c2272c6e7c98089. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,30 +1,6 @@
 import type { Field } from './field/type'
+import type { CreateHeadlessFormOptions } from './form'
 import type { JsfSchema } from './types'
-
-type DiskSizeUnit = 'Bytes' | 'KB' | 'MB'
-
-/**
- * @todo: Remove this.
- *
- * This utility only exists as an example of using V1 tests for V2 source.
- * It should not be tested, or even part of JSON Schema Form.
- */
-export function convertDiskSizeFromTo(
-  from: DiskSizeUnit,
-  to: DiskSizeUnit,
-): (value: number) => number {
-  const multipliers: Record<DiskSizeUnit, number> = {
-    Bytes: 1,
-    KB: 1024,
-    MB: 1024 * 1024,
-  }
-
-  return (value: number): number => {
-    const fromMultiplier = multipliers[from]
-    const toMultiplier = multipliers[to]
-    return (value * fromMultiplier) / toMultiplier
-  }
-}
 
 /**
  * Get a field from a list of fields by name.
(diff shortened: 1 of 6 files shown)


danielcardoso5:
[Internal pipeline](https://gitlab.com/remote-com/employ-starbase/dragon/-/pipelines/2703818958) 🟢

src/utils.ts
```diff
@@ -102,18 +78,40 @@ function getOptionIdentity(option: unknown): unknown {
...
+
+/**
+ * Warns (once) that a conditional branch introduces option(s) not present on the base field.
+ * This is only relevant while running with the legacy behavior (disallowNewConditionalOptions: false).
+ */
+function warnAboutNewConditionalOptions(): void {
+  if (!hasWarnedAboutNewConditionalOptions) {
+    hasWarnedAboutNewConditionalOptions = true
+    console.warn(
+      '[json-schema-form] A conditional branch introduces option(s) not present on the base field. '
```
sandrina-p: suggestion: Can we include the option name in the warning? It's much easier to debug.

sandrina-p (dismissed):
Approved! Just a comment on the warning msg, but besides that it looks great. Thanks for updating the MIGRATING.md file too ^.^

src/utils.ts
danielcardoso5: @sandrina-p I mentioned setting `disallowNewConditionalOptions: true` in the warning. Is that what you mean?

danielcardoso5: Ah, I understand now. Let me add it!

eshiota (approved):

PR #269, opened 2026-08-06
fix Use field defaults from schema in v1

danielcardoso5:
Fixes default values not being taken into account when initializing a form like v0 did. Supports passing default values for base fields, nested fields and group-array items.

With the schema from https://github.com/remoteoss/json-schema-form/issues/249:

| before | after |
|--------|--------|
| image | image | 

---

> [!NOTE]
> **Medium Risk**
> Changes initial form state and conditional visibility for all consumers relying on missing defaults; behavior differs from v0 for explicit falsy initial values.
> 
> **Overview**
> **Restores v0-style default seeding** so `createHeadlessForm` applies JSON Schema `default` values before the first field build, fixing conditionals (e.g. `if`/`then`) that depend on a defaulted field not resolving until after `handleValidation`.
> 
> A new **`fillDefaults`** helper walks the **base** schema: object properties, nested objects, and **existing** array items only. It sets `default` when the value is **`undefined`**—explicit falsy values like `null` are kept (documented as a v0 vs v1 migration difference). Defaults inside conditional sub-schemas are **not** applied. **`initialValues`** are not mutated.
> 
> **MIGRATING.md** adds a breaking-change note on default semantics; tests cover init-time conditionals, overrides, nested and group-array cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9f151df7ef847dfea7d5996c775e997d806ef7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -113,6 +113,16 @@ const { schema, warnings } = modify(schemaPet, {
 });
 ```
 
+### 7. **Default values are only applied when the initial value is `undefined`**
+
+Both versions fill a "missing" initial value with the schema's `default`, but they don't match on what counts as a "missing" initial value.
+
+- **v0** applied the default whenever the initial value was **falsy**, so `false`, `0`, `''` and `null` were all replaced by the default.
+- **v1** applies the default only when the initial value is **`undefined`**, so falsy values you explicitly pass are preserved.
+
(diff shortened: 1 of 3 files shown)


dragidavid (approved):

dragidavid:
Sorry for the delay @danielcardoso5

PR #271, opened 2026-08-19
fix: evaluate value-dependent items conditionals per array row

vermaxik:
## Summary

Conditionals (`if`/`then`/`else` in `allOf`) inside an array's `items` schema are
evaluated against an empty object instead of each row's value. Whichever branch
matches `{}` gets permanently merged into the shared `items` schema and deleted,
so rules with an `else` branch (or a negated `if`) validate wrongly for every
row. This PR pre-applies only constant conditionals inside `items` and leaves
value-dependent ones intact, so `validateSchema` evaluates them per row.

| Before    | After |
| -------- | ------- |
|  SCR-20260820-ksyz | SCR-20260820-ktnq   |

<details>

<summary>Json Schema</summary>
<pre>
{
  "title": "Per-row conditionals in nested arrays",
  "type": "object",
  "properties": {
    "discount_rules": {
      "type": "array",
      "title": "Discount rules",
      "x-jsf-presentation": { "inputType": "group-array", "addFieldText": "Add rule" },
      "items": {
        "type": "object",
        "title": "Rule",
        "x-jsf-order": ["name", "mode", "percent", "tiering"],
        "required": ["name", "mode"],
        "x-jsf-presentation": { "inputType": "fieldset" },
        "properties": {
          "name": {
            "type": "string",
            "title": "Name",
            "x-jsf-presentation": { "inputType": "text" }
          },
          "mode": {
            "type": "string",
            "title": "Mode",
            "default": "flat",
            "oneOf": [
              { "const": "flat", "title": "Flat" },
              { "const": "tiered", "title": "Tiered" }
            ],
            "x-jsf-presentation": { "inputType": "radio" }
          },
          "percent": {
            "type": ["string", "null"],
            "title": "Percent",
            "x-jsf-presentation": { "inputType": "text" }
          },
          "tiering": {
            "type": ["object", "null"],
            "title": "Tiering",
            "x-jsf-order": ["period", "tiers"],
            "x-jsf-presentation": { "inputType": "fieldset" },
            "properties": {
              "period": {
                "type": "string",
                "title": "Period",
                "oneOf": [
                  { "const": "monthly", "title": "Monthly" },
                  { "const": "yearly", "title": "Yearly" }
                ],
                "x-jsf-presentation": { "inputType": "select" }
              },
              "tiers": {
                "type": "array",
                "title": "Tiers",
                "x-jsf-presentation": { "inputType": "group-array", "addFieldText": "Add tier" },
                "items": {
                  "type": "object",
                  "title": "Tier",
                  "x-jsf-order": ["up_to", "percent"],
                  "required": ["percent"],
                  "x-jsf-presentation": { "inputType": "fieldset" },
                  "properties": {
                    "up_to": {
                      "type": "integer",
                      "title": "Up to",
                      "x-jsf-presentation": { "inputType": "number" }
                    },
                    "percent": {
                      "type": "string",
                      "title": "Percent",
                      "x-jsf-presentation": { "inputType": "text" }
                    }
                  }
                }
              }
            }
          }
        },
        "allOf": [
          {
            "if": {
              "properties": { "mode": { "const": "tiered" } },
              "required": ["mode"]
            },
            "then": {
              "required": ["tiering"],
              "properties": {
                "tiering": {
                  "type": "object",
                  "required": ["period", "tiers"],
                  "properties": {
                    "tiers": {
                      "minItems": 1,
                      "x-jsf-errorMessage": { "minItems": "Add at least one tier, or switch the mode to flat." }
                    }
                  }
                },
                "percent": {
                  "maxLength": 0,
                  "x-jsf-errorMessage": { "maxLength": "Not used for tiered rules. Clear it or switch the mode to flat." }
                }
              }
            },
            "else": {
              "required": ["percent"],
              "properties": {
                "percent": { "type": "string" },
                "tiering": {
                  "properties": {
                    "tiers": {
                      "maxItems": 0,
                      "x-jsf-errorMessage": { "maxItems": "Tiers only apply to tiered rules. Remove them or switch the mode." }
                    }
                  }
                }
              }
            }
          }
        ]
      }
    }
  }
}
</pre>
</details>

## Changes Made

`calculateFinalSchema` pre-applies conditional rules by merging the matching
branch into the schema and deleting the branch. That is correct at the root,
where `values` is the real form value, but for `items` the existing workaround
in `applySchemaRules` passed a hardcoded `{}`:

- a rule with an `else` branch had the `else` baked in for all rows (an `if`
  that requires a field never matches `{}`), producing wrong validation even
  for rows where the condition is true;
- a negated `if` always matches `{}`, baking the `then` in for all rows;
- because the branch is deleted after merging, the per-row conditional
  evaluation in `validateCondition` (which is correct) never saw it.

The change threads a `constantIfsOnly` flag through `applySchemaRules`. Inside
`items`, only conditionals whose `if` is a boolean (`if: true` / `if: false`)
are pre-applied — their branch is row-independent, which is what the original
workaround supported (schema-driven visibility inside items, covered by the
existing "with constant logic" tests). Value-dependent rules keep their
`then`/`else` and are evaluated per item with the row's actual value.

Known limitation, unchanged by this PR: per-row FIELD state (visibility,
required flags, titles) is still not representable, since all rows of a
group-array share a single fields array. This PR fixes validation only; the
`describe.skip('with logic based on answers')` for group-array field visibility
stays skipped.

---

> [!NOTE]
> **Medium Risk**
> Touches core schema mutation in `applySchemaRules`, which affects validation and field derivation for nested arrays; behavior change is intentional but could surface edge cases in complex conditional schemas.
> 
> **Overview**
> Fixes **wrong validation for array row schemas** when `items` uses value-dependent `if`/`then`/`else` rules. Previously `applySchemaRules` ran against `{}` for shared `items` schemas, permanently merging one branch and breaking `else`/negated `if` for every row.
> 
> The change adds **`constantIfsOnly`** (via `shouldProcessRule`) so only **`if: true` / `if: false`** conditionals are pre-merged into `items`; value-dependent rules stay on the schema and are evaluated **per row** during validation. Constant conditionals on nested objects and array items still drive shared field metadata (e.g. visibility/required).
> 
> New array tests cover `else` branches, negated `if`, nested constant+value-dependent rules, and `if: true` on objects and items. **Per-row field UI mutations** for group-arrays remain out of scope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1d4962336fed905500858bdfaf95663263f556f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/mutations.ts
+++ b/src/mutations.ts
@@ -125,13 +125,13 @@ function applySchemaRules(
   for (const { rule, matches } of conditionalRules) {
     // If the rule matches, process the then branch
     if (matches && rule.then) {
-      processBranch(schema, values, rule.then, options, jsonLogicContext)
+      processBranch(schema, values, rule.then, options, jsonLogicContext, constantIfsOnly)
       // Delete the then branch to avoid processing it again when validating the schema
       delete rule.then
     }
     // If the rule doesn't match, process the else branch
     else if (!matches && rule.else) {
-      processBranch(schema, values, rule.else, options, jsonLogicContext)
+      processBranch(schema, values, rule.else, options, jsonLogicContext, constantIfsOnly)
       // Delete the else branch to avoid processing it again when validating the schema
       delete rule.else
     }
@@ -174,11 +174,12 @@ function applySchemaRules(
  * @param branch - The branch (schema representing and then/else) to process
  * @param options - Validation options
  * @param jsonLogicContext - JSON Logic context
+ * @param constantIfsOnly - When true, only constant (boolean) conditionals are pre-applied
  */
-function processBranch(schema: JsfObjectSchema, values: SchemaValue, branch: JsfSchema, options: CreateHeadlessFormOptions = {}, jsonLogicContext: JsonLogicContext | undefined) {
+function processBranch(schema: JsfObjectSchema, values: SchemaValue, branch: JsfSchema, options: CreateHeadlessFormOptions = {}, jsonLogicContext: JsonLogicContext | undefined, constantIfsOnly: boolean = false) {
   const branchSchema = branch as JsfObjectSchema
 
-  applySchemaRules(branchSchema, values, options, jsonLogicContext)
+  applySchemaRules(branchSchema, values, options, jsonLogicContext, constantIfsOnly)
   mergeSchemaBranch(schema, branchSchema, options)
 }
 
--- a/src/mutations.ts
+++ b/src/mutations.ts
@@ -89,15 +89,19 @@ function applySchemaRules(
   values: SchemaValue = {},
   options: CreateHeadlessFormOptions = {},
   jsonLogicContext: JsonLogicContext | undefined,
+  constantIfsOnly: boolean = false,
 ) {
   if (!isObjectValue(values)) {
     return
   }
 
+  const shouldProcessRule = (ifNode: JsfSchema | undefined): boolean =>
+    typeof ifNode !== 'undefined' && (!constantIfsOnly || typeof ifNode === 'boolean')
+
   const conditionalRules: { rule: NonBooleanJsfSchema, matches: boolean }[] = []
 
   // If the schema has an if property, evaluate it and add it to the conditional rules array
-  if (typeof schema.if !== 'undefined') {
+  if (shouldProcessRule(schema.if)) {
     conditionalRules.push(evaluateConditional(values, schema, schema, options, jsonLogicContext))
   }
 
@@ -105,7 +109,7 @@ function applySchemaRules(
   const allOf = schema.allOf ?? []
   const jsonLogicAllOf = schema['x-jsf-logic']?.allOf ?? [];
 
-  [...allOf, ...jsonLogicAllOf].filter((rule: JsfSchema) => typeof rule.if !== 'undefined').forEach((rule) => {
+  [...allOf, ...jsonLogicAllOf].filter((rule: JsfSchema) => shouldProcessRule(rule.if)).forEach((rule) => {
     const result = evaluateConditional(values, schema, rule, options, jsonLogicContext)
     conditionalRules.push(result)
   })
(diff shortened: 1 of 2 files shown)


src/mutations.ts
```diff
@@ -89,23 +89,27 @@ function applySchemaRules(
   values: SchemaValue = {},
   options: CreateHeadlessFormOptions = {},
   jsonLogicContext: JsonLogicContext | undefined,
+  constantIfsOnly: boolean = false,
 ) {
   if (!isObjectValue(values)) {
     return
   }
 
+  const shouldProcessRule = (ifNode: JsfSchema | undefined): boolean =>
```
eshiota: Nit: this could be moved outside of the `applySchemaRules`'s scope by injecting `constantIfsOnly` as an argument

test/fields/array.test.ts
```diff
@@ -701,6 +701,122 @@ describe('buildFieldArray', () => {
...
+      })
+
+      // A banded row must not be forced into the else branch requirements
+      expect(form.handleValidation({
+        rules: [{ name: 'banded', banding: { period: 'daily', bands: [{ rate: '2.0' }] } }],
+      }).formErrors).toEqual(undefined)
+    })
+
+    it('evaluates negated conditionals per array item', () => {
+      // A negated `if` matches the empty object, so the old pre-processing baked
```
eshiota: Nit: we don't need most of these comments, the code should be self-explanatory

test/fields/array.test.ts
eshiota, line 1: Could we add additional tests that cover some of the changes directly?

- An `if: true` clause inside an `allOf` at the level of a property with `type: 'object'` 

```
address: {
  'type': 'object',
  'x-jsf-presentation': { inputType: 'fieldset' },
  'properties': { city: {...}, zip: {...} },
  'allOf': [{
    if: true,
    then: { required: ['city'], properties: { zip: false } },
    else: { required: ['zip'], properties: { city: false } },
  }],
}
```

- An `if: true` clause directly at the level of a property with `type: 'object'`

```
items: {
  type: 'object',
  properties: { a: {...}, b: {...} },
  if: true,
  then: { required: ['a'] },
  else: { required: ['b'] },
}
```

- The same as above, but with a property with `items` as an array

eshiota (changes requested):
Everything looks solid! I left two nits as comments, and a request to add additional tests 😄

test/fields/array.test.ts
vermaxik: added in https://github.com/remoteoss/json-schema-form/pull/271/changes/590d4f7073c5393f88894aa59b046c8211329656

src/mutations.ts
vermaxik: changed in https://github.com/remoteoss/json-schema-form/pull/271/changes/590d4f7073c5393f88894aa59b046c8211329656

test/fields/array.test.ts
vermaxik: removed in https://github.com/remoteoss/json-schema-form/pull/271/changes/590d4f7073c5393f88894aa59b046c8211329656

vermaxik:
> Everything looks solid! I left two nits as comments, and a request to add additional tests 😄

thanks for the review, addressed feedback in https://github.com/remoteoss/json-schema-form/pull/271/changes/590d4f7073c5393f88894aa59b046c8211329656

src/mutations.ts
```diff
@@ -131,23 +142,25 @@ function applySchemaRules(
...
-          * This is not ideal, but it's better than the previous situation where the rules were not applied at all.
+          * Only constant conditionals (`if: true` / `if: false`) are pre-applied here —
+          * their branch is the same for every row, so baking them is safe and keeps
+          * schema-driven visibility inside items working. Value-dependent rules are
+          * left intact so `validateSchema` (via `validateCondition`) evaluates them per
+          * item with the row's actual value. Per-row FIELD mutations (visibility /
+          * required flags) remain unsupported for group-array items.
           */
-          applySchemaRules(propertySchema.items as JsfObjectSchema, {}, options, jsonLogicContext)
+          applySchemaRules(propertySchema.items as JsfObjectSchema, {}, options, jsonLogicContext, true)
```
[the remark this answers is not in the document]
vermaxik: fixed in https://github.com/remoteoss/json-schema-form/pull/271/changes/4cff9448cfc5dafe189562286ed3ad83c35c531e

eshiota (approved):

